PR #243, opened 2021-04-06
move from respec-common to respect

airmisdata:


davidcarlisle:
@airmisdata hi I don't follow respec updates that closely or recall why I used the -common version originally, would it be possible for you to add a comment on the effect of changing version here?

airmisdata:
Sure @davidcarlisle  ! respec-common is now deprecated https://github.com/w3c/respec/wiki/respec-w3c-common-migration-guide

davidcarlisle:
@airmisdata  ah thanks that seems clear enough:-)

davidcarlisle:
@airmisdata  I'll do the notes-on mathml (which is currently more or less a copy of this) to match, thanks.

PR #270, opened 2022-03-01
restructure fundamentals chapter

davidcarlisle:
Restructuring fundamentals chapter (description to be filled in when it is clearr what chnages have been made)
this is currently a draft PR, not for merging


davidcarlisle:
This re-write is far from complete but I am going to ask for review and merge at this stage to keep the diffs relatively small.
(the actual diff here is large due to cleanup and re-indenting of the file which was full of artifacts of the initial conversion to respec from the original xml sources.

The main changes are

* A move to using standard respec/specref  refence ids so we can lose the localbiblio javascript setting in the respec config, see https://www.specref.org/?q=MathML
* removal of all local specification of color syntax (defering to css3-color)
* removal of suggestion of using xslt for class and style and directly referencing css (mathml3 said the syntax of these attributes was unspecified)
* Marking which global attributes are inherited from core and which are not (he actual marking of that is a bit minimal currently but it gets the data into this file which should simplify future edits)

NSoiffer (approved):
I can't see anything wrong here, but then again, I can't see the effect of the changes...

I'm approving this so the group (and I) can see the changes.

PR #275, opened 2022-03-05
re-instate the index of elements

davidcarlisle:
The MathML3 spec had an appendix with an index of elements (and a separate index of attributes) 

https://www.w3.org/TR/MathML3/appendixi.html

This was dropped in the conversion to respec as the pre-processing to generate the index didn't really fit the respec processing model.

This PR re-instates the index, generated this time via javascript in a respec postprocess config option, so it has some hit on the respec processing time  but may possibly be useful.

Adding as a PR so people can decide if they want it (basically it's a small addition, one new stub html fil to make the appendix and <50 lines of javascript, but it does generate a relatively big appendix and have some time cost on every build.

This is more or less independent of any other edits we do so no rush to merge, it could be merged later if it's not clearly a good thing.

The resulting index currently looks like

image


NSoiffer:
It looks good to me, but since it affects everyone in the group, I want to bring it up at the meeting.

PR #277, opened 2022-03-08
compressed display of authors and WG members

davidcarlisle:
The contributors lists in appendix E take up a lot of space (even before we add any new list for MathML4).  This is a fairly mild re-arrangement keeping all the list text but displaying the list of names inline rather than a description list, with affiliations just in `title` attribute popups.

so taking just one of the lists 

image

becomes

image

We could compress further eg merge the lists for mathml 1/2/3 since most people appear at least twice, but starting with small steps.

I suspect most specs don't now list full WG membership just a subset of "authors" or smaller subset of "editors" but previously we have conciously listed everyone not least because some contributors for whom English isn't their first language have contributed in ways other than actually editing the text, and would not be recognised 

I pinged the co-chairs for review as this is a policy thing rather than anything else, but anyone feel free to comment.

### DIFF
--- a/src/contributors.html
+++ b/src/contributors.html
@@ -19,94 +19,66 @@
 
   <p>Participants in the Working Group responsible for MathML 3.0 have been:
   </p>
-  <dl>
 
-   <dt>Ron Ausbrooks</dt>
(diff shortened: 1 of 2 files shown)


NSoiffer:
WG agreed.

Issue #266, opened 2022-02-24
Add a new chapter explaining core/presentation


NSoiffer:
We need to explain the difference between core and full spec. The group felt that a new chapter early on is appropriate.

The feeling is this is should be written as "full is an addition to core".

It should say that a web platform implementation needs to follow core's normative statements.

Are they free to add the additional attributes that are in full? E.g., should core have negative tests like saying mfenced doesn't work as in full.

An analogy is language design where there is a core and a standard library. Maybe use that model.

@brucemiller will take the first crack

NSoiffer:
Bruce to put this in intro section near presentation vs content.

Will also clean out non-normative.

PR #283, opened 2022-03-10
First draft of drawing contrast of Full with Core

brucemiller:
Rewrite the "Overview" section within Introduction Chapter, to cast MathML Full as an extension of MathML Core.

This PR addresses #266

### DIFF
--- a/src/introduction.html
+++ b/src/introduction.html
@@ -272,15 +272,16 @@
       
       <p>
          The specification of MathML is developed in two layers.
-         <span>MathML Core</span> covers (most of) Presentation Markup,
-         with the focus being the display of mathematics in web browsers.
+         <span>MathML Core</span>([[mathml-core]]) covers (most of) Presentation Markup,
+         with the focus being the precise details of displaying mathematics in web browsers.
          <span>MathML Full</span>, this specification, extends <span>MathML Core</span>
          primarily by defining Content MathML, in <a href="#contm"></a>.
          It also defines extensions to Presentation MathML, discussed in <a href="#presm"></a>.
          These extensions consist of additional attributes, elements or
          enhanced syntax of attributes.
          These are defined for compatibility with legacy MathML,
-         as well as to cover more complex notations.
+         as well as to cover <a href="#presm_linebreaking"></a>, <a href="#presm_elementary"></a>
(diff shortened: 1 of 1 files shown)


davidcarlisle (approved):

src/introduction.html
```diff
@@ -260,33 +260,46 @@ <h4 id="intro_goals">Design Goals of MathML</h4>
...
+         The specification of MathML is developed in two layers.
+         <span>MathML Core</span> covers (most of) Presentation Markup,
+         with the focus being the display of mathematics in web browsers.
+         <span>MathML Full</span>, this specification, extends <span>MathML Core</span>
+         primarily by defining Content MathML, in <a href="#contm"></a>.
+         It also defines extensions to Presentation MathML, discussed in <a href="#presm"></a>.
+         These extensions consist of additional attributes, elements or
+         enhanced syntax of attributes.
+         These are defined for compatibility with legacy MathML,
+         as well as to cover more complex notations.
```
dginev: I wonder if `less common` or `harder to implement` is a better emphasis than `more complex` here, as one can compose together MathML Core to cover some really complex notations already.

I worry that the sentence may sound discouraging that e.g. MathML Core couldn't/shouldn't be used for higher mathematics.

davidcarlisle: @dginev  we could say something like ... 

```
....
as well as to cover <a>Linebreaking of Expressions</a>, <a>Elementary Math</a>
and other aspects not included in level 1 of [[[mathml-core]]] 
```

So that it's less subjective, and just states that the things in the referenced sections are not in the referenced spec?

NSoiffer:
>       The specification of MathML is developed in two layers.
>          <span>MathML Core</span> covers (most of) Presentation Markup,
>          with the focus being the display of mathematics in web browsers.

I'd like to emphasis that Core is not just display math in browsers, but the exact specification of the display. Perhaps "...the details of displaying mathematics....".

src/introduction.html
brucemiller: sounds good; done!

brucemiller:
good point; done!

NSoiffer:
Good enough for now. Approving.

PR #286, opened 2022-03-12
A start at refactoring chapter 5 for intent

davidcarlisle:
This is a very "early draft" PR I don't expect any of the text to survive future edits, but asking for review/merge anyway to firm up the plans for this chapter,

The main changes here

* chapter name
* radically cut down `<semanics>` examples especially those using prefixed OpenMath that won't work in HTML (the version of this in the Notes-on-MathML document can keep these and have more examples comparing with intent markup.)
* Have an initial section describing `intent=`. currently this has a grammar taken from comment in the "Unifying intent"  github issue and some text stolen from the mathml-refresh semanics-mini paper (which was actually describing as slightly different proposal)

Any discussion of intent here is going to need to cross reference with discussion of intent in chapter 3 (there has to be some discussion there if only to say the attribute is allowed, so wording here is necessarily speculative, also the exact grammar was never finally agreed.

But asking for review on the basic plan of making this chapter describe both `intent=` and `<semantics>`  as being markup to solve the same basic issue, with `<semanics>` being more general and more verbose, and `intent` being specialised to annotating presentation (and core) mathml.


src/mixing.html
```diff
@@ -1,31 +1,112 @@
 <section data-max-toc="3">
- <h2 id="mixing">Mixing Markup Languages for Mathematical Expressions</h2>
+ <h2 id="mixing">Semantic Annotations</h2>
```
brucemiller: Nice & brief, but annotations like svg or TeX aren't necessarily "semantic".

src/mixing.html
```diff
@@ -1,31 +1,112 @@
...
+ markup and content markup to be combined in several different ways
+ and for MathML to embed markup from other vocabularies such as
+ [[OpenMath2004]] or [[SVG]]. This general mechanism is described
+ below in <a href="#mixing_semantic_annotations"></a> but first we describe a new
+ feature added in MathML&#160;4 to faciliate annotating Presentation
+ MathML with the <q>intent</q> or semantic meaning of the term.
  </p>
 
+ <section>
+  <h3 id="mixing_intent">Semantic Intent for Presentation MathML</h3>
```
brucemiller: Is it worth pondering, for a brief moment, that `intent` at least as we've defined it so far, could apply equally well to content, and my conceivably be useful. (though of course defaulting patterns would be different)

src/mixing.html
```diff
@@ -1,31 +1,112 @@
...
+  &lt;mi>x&lt;/mi>
+  &lt;mo>'&lt;/mo>
+&lt;/msup>
+</pre>
+
+ </section>
+ </section>
+
+ 
  <section>
```
brucemiller, line 106: Nice set of examples!

brucemiller (approved):
This is quite acceptable, given it will likely change as we go.  But a few comments for the record.

src/mixing.html
dginev: Note on "semantic intent"

A big part of the discussion that led to the `intent` term, for the attribute name, was to very carefully avoid the baggage (and associations) which the word "semantic" brought along. Especially given the spec has both a Content MathML dialect, *and* a `<semantics>` element.

So far I have been using the phrase "author intent" when "intent" needed to be qualified, as well as "encyclopedic intent" when discussing the way in which I have been choosing the values in the lists of sample values.

But I would be equally happy with "accessible intent" or the simple and neutral "intent attribute", in such a heading title followed by "for Presentation MathML".

Just trying to keep the reader's mind free of prior bias, when first encountering the new attribute.

src/mixing.html
dginev: I would second Bruce here.

"Foreign markup annotations" could be a neutral phrase. Just to offer one example.

davidcarlisle: yes I think I'll change to `Annotating MathML`, don't need to say at this point point why or what we are annotating with

src/mixing.html
davidcarlisle: I had a similar thought while discussing wih @samdooley  on last week's call the strict content rewrite rules. I'll open an issue on that and add an issue link to this section.

davidcarlisle: There was some discussion of this in #289 with a decision to park this at least for this MathML4 iteration, so marking this as resolved.

PR #288, opened 2022-03-12
follow core, defer to css for length syntax

davidcarlisle:
Second iteration of update to mathml fundamentals chapter
this makes all lengths valid in core valid with the same meaning in full.

Notably this allows additional css units and re-interpets px as a fixed size notional 96dpi not an implemention defined based on the hardware pixels (although I don't call that out here)


NSoiffer (approved):
Looks ok to me.

Issue #291, opened 2022-03-14
Drop the Glossary?


davidcarlisle:
I suggest we drop the Glossary, https://w3c.github.io/mathml/#glossary from the normative spec.

If there are any terms that need defining we can define them inline, marked with `<dfn>` and then any later uses of the term can be marked with ReSpec `[=thisterm=]` which will link back to the definition, so the separate glossary list isn't so useful and risks being in conflict with normative definitions elsewhere in the spec or in mathml core.

NSoiffer:
Good call. I don't remember ever looking at it. More importantly, I checked the current SVG, HTML, and a few EPUB drafts and none of them had one. WCAG does have one though. Overall, they don't seem to be very common AFAIK.

PR #292, opened 2022-03-14
delete glossary for issue #291

davidcarlisle:
also adjust axis references to use `<dfn>`  (axis was the only glossary entry referenced explicitly)


brucemiller:
Seems reasonable.

NSoiffer:
WG agreed

PR #293, opened 2022-03-14
delete macros and mode attributes, reference legacy schema

davidcarlisle:
Just a small change but we might want to decide a policy on deprecated things generally

here I have taken a hard line and removed all description of mode and macros, treating them as any unknown invalid attribute, except reference the mathml-legacy schema in appendix A which makes them valid again.


NSoiffer:
Let's discuss this at the meeting and see what the group thinks.

I believe in the past the policy was that deprecated items hang around as deprecated for one (major) new version of the spec and then go away. So anything deprecated in V2 should go away. But I'm not certain that was our policy.

davidcarlisle:
On Mon, 14 Mar 2022 at 21:21, NSoiffer ***@***.***> wrote:

> Let's discuss this at the meeting and see what the group thinks.
>
> I believe in the past the policy was that deprecated items hang around as
> deprecated for one (major) new version of the spec and then go away. So
> anything deprecated in V2 should go away. But I'm not certain that was our
> policy.
>

we chickened out in mathml3, mode was deprecated in mathml2 and left as
deprecated in mathml3, if we leave it in again we may as well just say it's
in full and not deprecated, it as it'll never go.

Currently all (or most, I haven't done a complete sweep to check)
deprecated things are not in the appendix A schema, except the final
mathml-legacy schema that puts them all back.   I think we should do that
anyway even if we describe behaviour for the deprecated features.
Especially in an html world declaring them as invalid but with defined
behaviour seems quite natural. But for mode that has been deprecated for 20
years, and macros never had any defined behaviour even before that, I'd say
to let them go and make them have undefined behaviour.

But yes the PR was really for the meeting agenda, no need to merge before
that.

The easiest way to see (what appendix A thinks are) the deprecated items is
to look out for the bold terms in the schema at

https://w3c.github.io/mathml/#parsing_rnc_legacy

NSoiffer:
WG agreed

PR #295, opened 2022-03-17
Make Introduction informative (non-normative)

### LINKED ISSUE #263: Introduction
The intro needs to move to informative

It should follow other spec intro.

Steve Noble to do work

### LINKED ISSUE #266: Add a new chapter explaining core/presentation

brucemiller:
As it says.

Fixes #263
Fixes #266

(unless we want to keep those issues open for further/other discussion)

### DIFF
--- a/src/introduction.html
+++ b/src/introduction.html
@@ -1,4 +1,4 @@
-<section>
+<section class="informative">
  <h2 id="intro">Introduction</h2>
  
  <div class="issue" data-number="263"></div>


davidcarlisle (approved):

davidcarlisle:
merging this closed both issues.  263 should clearly be fixed, I doubt we have the final version of the relationship to core, but we'll need to come back to that after chapters 2 and 5 updated, it will probably be clearer to let 266 close now and open any new issues on the draft spec later.

PR #297, opened 2022-03-19
compress length description

davidcarlisle:
Following a similar path to color, this re-interprets mathml lengths as a modification from the css3 `length-percentage` (the only syntax allowed in core)

as it no longer "builds up" from definitions of digit and hexdigit these definitions are removed (which will leave some broken links from the presentation mathml syntax tables that will need fixing on a second sweep through).

added are named lengths, unitless lengths and mpadded `height`, `depth`, `width` pseudo-units.

Not added are magic interpretations of + and - signs in lengths (I think with core defering to css here we need to do the same 

That is, MathML3 said of mpadded width="+3em"

> If the value of a size attribute begins with a + or - sign, it specifies an increment or decrement to the corresponding dimension by the following length value.

But in mathml-core it is the same as width=3em and I think we have to follow core here.

This doesn't directly affect the PR which is only defining the syntax, +3em is legal in either case, but commenting here as this would be a technically breaking change.

We could consider adding new attributes eg change-width="3em" which would be a lot easier to understand than width="+3em"

### DIFF
--- a/src/fundamentals.html
+++ b/src/fundamentals.html
@@ -206,12 +206,12 @@
       </tr>
 
       <tr>
-       <td><dfn><em>length</em></dfn></td>
+       <td id="type_length"><dfn><em>length</em></dfn></td>
        <td>a length, as explained below, <a href="#fund_units"></a></td>
       </tr>
        
       <tr>
-       <td><dfn><em>namedlength</em></dfn></td>
+       <td id="type_namedspace"><dfn><em>namedspace</em></dfn></td>
        <td>a named length, as explained below, <a href="#fund_units"></a></td>
       </tr>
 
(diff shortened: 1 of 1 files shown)


NSoiffer (commented):
I'm seeing three link errors:
```
<dfn><em>positive-integer</em></dfn>
```
Also `integer` and `unsigned-number`.

I'm not keen on the new format for named lengths. How about a two column table with the first column being the positive lengths and the second one the negative lengths? It wouldn't be much longer but it would be much clearer and highlight the symmetry.

Do you think we should deprecate omitting the explicit unit (unless it is 0) for compatibility with the rest of the ecosystem?

davidcarlisle:
@NSoiffer  MathML3 already said of unitless lengths

 https://www.w3.org/TR/MathML3/chapter2.html#type.length

>  A number without a unit is intepreted as a multiple of the reference value. This form is primarily for backward compatibility and should be avoided, prefering explicit units for clarity. 

So they were effectively deprecated in MathML3 (although we didn't use that word).

Especially if we decide to drop the mpadded `+3em` syntax it would make sense to drop these as well from the main schema/spec and just have them in the explicit legacy schema.

Perhaps we should pause this PR and open issues for those changes and then come back.

Yes there are some link warnings due to  the restructuring of the terminals (digit and integer were used to build up the length syntax but here I'm starting from css length-percentage. I think it's easy to do a sweep over the references to fix that up but I didn't want to do that until the base change is agreed as it's a lot of small changes (and would probably make merge conflicts elsewhere unless we coordinate the order of changes.)

brucemiller:
Regarding "change-width", one could also consider extending `rspace`,`lspace` to more (or all) elements.  I suspect such a late change to Core would be unpopular, however.

PR #298, opened 2022-03-19
update operator dictionary description to match current version

davidcarlisle:
Deleted all the text describing the MathML3 operator dictionary presentation, added some text describing the new presentations.

Also as requested on the last call, added reference to machine readable `unicode.xml` source for the data.

I used the phrase "suitable browsers" when describing the sortable table interaction in the second view and the title popups in the first. Not sure if that is an acceptable phrase or if there is a more inclusive way of describing the interaction that covers non visual  renderings of this data.

### DIFF
--- a/src/operator-dict.html
+++ b/src/operator-dict.html
@@ -1,174 +1,52 @@
 <section class="appendix">
  <h2 id="oper-dict">Operator Dictionary</h2>
 
- <div class="issue" data-number="87"></div>
- <div class="issue" data-number="274"></div>
+ <!--<div class="issue" data-number="87"></div>-->
+ <!--<div class="issue" data-number="274"></div>-->
 
  <p>The following table gives the suggested dictionary of rendering
- properties for operators, fences, separators, and accents in MathML, all of
- which are represented by <code class="element">mo</code> elements. For brevity,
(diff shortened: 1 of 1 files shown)


src/operator-dict.html
```diff
@@ -1,174 +1,52 @@
...
+ elements. For brevity, all such elements will be called
+ simply <span>&#x201c;operators&#x201d;</span> in this Appendix. Note
+ that implementations of [[MathML-Core]] will use these values as
+ normative definitions of the default operator spacing,</p>
 
  <section>
   <h3 id="oper-dict_index">Indexing of the operator dictionary</h3>
 
-  <p>Note that the dictionary is indexed not just by the element
+  <p> Th dictionary is indexed not just by the element
```
brucemiller: The

NSoiffer (changes requested):
Small typo:  " Th dictionary"

Otherwise looks good.

davidcarlisle:
> Small typo: " Th dictionary"
> 
> Otherwise looks good.

Bruce flagged this, fixed in 421159481617e2d1a4d39cc1105d043a0464ed15 I think?



>

Issue #258, opened 2021-01-12
remove media type registrations


samuelweiler:
These were done in the level 3 spec and do not need to be repeated here.  It would be appropriate to point out that this version (re-)uses these media types, which were previously registered, but we don't need to repeat the template again here.  (and, indeed, repeating it could lead to confusion).

davidcarlisle:
@samuelweiler  @NSoiffer @polx  Just referencing some discussion at

https://github.com/w3c/mathml/pull/299#issuecomment-1073642575

as PR comments more easily lost than issue links,

It seems other W3C media registrations (notably SVG) effectively end up referencing the current version of the specification which seems preferable in the long run than referencing a specification that has been superseded.

davidcarlisle:
@samuelweiler  @NSoiffer @polx  Just referencing some discussion at

https://github.com/w3c/mathml/pull/299#issuecomment-1073642575

as PR comments more easily lost than issue links,

It seems other W3C media registrations (notably SVG) effectively end up referencing the current version of the specification which seems preferable in the long run than referencing a specification that has been superseded.

davidcarlisle:
Referencing obsolete versions does have some issues though, for example XSLT 3 just references the XSLT2 media type registration

https://www.w3.org/TR/xslt-30/#xslt-mime-definition

But the URL given there (and at the IANA registraion for application/xslt-xml) is

https://www.w3.org/TR/2007/REC-xslt20-20070123/#media-type-registration

But this is an obsoleted document that produces the dialog

![image](https://user-images.githubusercontent.com/1268738/159981730-e1b33549-f227-4615-9eef-e25393010e99.png)

It would be good if we could arrange with IANA to give a stable document that just has the registration without being tied to a versioned mathml x.y rec, then we could reference that from MathML 4, ie change the MathML3 links listed at

https://www.iana.org/assignments/media-types/media-types.xhtml#application

polx:
Request to [IETF-types sent](https://mailarchive.ietf.org/arch/msg/media-types/tnsgLJ4zN7etCAA72NWdaDIbgUo/).

polx:
An [answer came quickly](https://mailarchive.ietf.org/arch/msg/media-types/e9-odQ-usqIwpRBrFPVSmWSyqrU/) by Martin Dürst: it is all possible to update and we apply the same process as the registration. That means that we can keep the appendix and a (maybe reduced) part of chapter 6.

davidcarlisle:
[...]

PR #299, opened 2022-03-20
defer to mathml3 media type registration for issue #258

davidcarlisle:
As suggested in issue #258 this deletes the media type registration text. Instead referencing the MathML3 version and just leaving a brief description of the use. I moved the appendix later at the same time.

@polx I tried to add you as a reviewer but your name didn't show, is your gh account not logged with w3c?


NSoiffer (commented):
Maybe it is too late at night for me, but I'm a little baffled by the plan. Why is it bad to list them in MathML 4?

To me, it seems strange/inconsistent to say "it is repetitious to repeat their definition in MathML 4 because they were defined in MathML 3". A huge percent of MathML 4 is defined in MathML 3, but the point of a new spec is list all aspects of the current version of the spec. If the Media types were in their own spec, then it would make sense to reference that spec, but they are in the MathML 3 spec which is superseded by MathML 4.

davidcarlisle:
@NSoiffer  yes I'd prefer to leave it in but the PR is there if this is the policy and we have to drop them, I think the issue is that the mathml3 URL is listed at iana and changing that is more difficult than not changing it.

Specifically

https://www.iana.org/assignments/media-types/media-types.xhtml#application

Lists the Mathml3 URL  that seems to be unfortunate and anomalous, perhaps we can change that

Other W3C registrations don't have that issue:

SVG lists 
https://www.w3.org/TR/SVG/mimereg.html

which is actually a redirect of

https://www.w3.org/TR/SVG2/mimereg.html

appendix J of the SVG2 CR not the original appendix P of SVG1.1

https://www.w3.org/TR/2011/REC-SVG11-20110816/mimereg.html

although (as we were proposing for mathml, the actual text is the same)





PNG  just lists the working group
https://www.iana.org/assignments/media-types/media-types.xhtml#PNG_Working_Group

application/html just lists Robin Berjon and the text of the registration uses the unversioned URL

Published specification :
http://www.w3.org/TR/html

so effectively allows in-place updates.

polx:
Conceptually, my problem is the naming of "level" that was introduced in #258 by @samuelweiler : I've seen this naming in concrete action for CSS and tend to believe that CSS has such a broad implementation basis that most browsers support several levels.

I don't believe this is ever going to be possible for MathML and would thus rather hope we can keep the registration traces in form of the appendix, e.g. stating that this is a repeat (for an appendix, that's ok). You re right @davidcarlisle in your comment, that we should have taken an upgradable URL as pointer. We could do an update IETF, I don't think that this is difficult, even if just to update the URL. All this would happen when we go in candidate-recommendation or a similarly finished status.

PR #300, opened 2022-03-21
278 collect pragmatic to strict

samdooley:
First attempt to collect the pragmatic-to-strict rewrite rules into section 4.6.  Most of the edits from sections 4.3 and 4.4 should be in good shape, but there are random references to the transformation algorithm in other sections that will need to be reviewed if this works for folks.

I recommend starting with https://samdooley.github.io/mathml/#contm_p2s and looking back at other sections as needed.  Feel free to suggest other reviewers.

### DIFF
--- a/src/contm-new.html
+++ b/src/contm-new.html
@@ -838,239 +838,9 @@
      a general presentation construct (see <a href="#presm_summary"></a>),
      which is used for rendering (see <a href="#contm_rendering"></a>).</p>
 
-     <p><em>Mapping to Strict Content MathML</em></p>
-
      <p>If a <code class="attribute">base</code> attribute is present, it specifies the base used for the digit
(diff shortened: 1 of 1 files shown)


davidcarlisle:
I think this looks good and could probably be merged even if there ar more edits to do.

As I commented somwhere I did wonder whether there should be links in the synax tables to the approproate rules, but looking now, maybe not, there could be several rules depending on the usage so the syntax table might start looking complicated. the OM symbols entry tells you which csymbol to use in whichever rule is found. It looks a bit lost without the "strict" parts but has to be with the symbol entry in 4.3/4.4 I think so OK.


I suspect that the whole of 4.6 could now be moved to a normative appendix (and raised one level in the section hierarchy)

davidcarlisle (approved):

PR #301, opened 2022-03-22
Davidc/chars

davidcarlisle:
This PR moves the opening sections of the characters chapter to the Notes document, updating the text for Unicode 14, and MathML Core.

The removed sections are visible at

https://w3c.github.io/mathml-docs/notes-on-mathml/#notes-on-unicode-in-mathml

Some of the remaining sections may need WG discussion and parts stay in the normative spec, but I think these are all non controversial background information sections that can be moved easily.


NSoiffer:
Looks good. Hopefully the media type issue can be resolved in a better manner.

Issue #253, opened 2021-11-12
n-ary Intent discussion


dginev:
I'd like to move the n-ary operation discussion from #252 in a dedicated issue, also so that we collect and discuss a few diverse examples.

## Examples

Here are three I picked from the MathML already in "the wild". I will be attaching screenshots since github has no supported way to render math expressions at the moment.

### n-ary arrows

The first formula on [this tungsteno page](https://www.tungsteno.io/post/def-de_rham_cohomology_groups/), referred to as a "chain of maps", using 8 right arrows, each decorated from above with a `d` :

<img height="40" src="https://user-images.githubusercontent.com/348975/141469302-3236327f-b15f-4f60-be32-44f46a7871d2.png">

the presentation of each operator is:
```xml
<mover>
  <mo stretchy="true" minsize="3.0em">→</mo>
  <mpadded width="+0.6em" lspace="0.3em">
    <mi mathvariant="normal">d</mi>
  </mpadded>
</mover>
```

* the second display formula in that page is similar, but the embellishment is *different* on each arrow.
* another interesting bit to notice is that since the MathML is katex-generated, there is no auxiliary mrow structure at all. The top-level of the expression has a single mrow, which contains ~40 direct children, one for each horizontal piece. The parenthetical Ω applications are not grouped, and there are no invisible unicode characters anywhere in the expression.

Equation (2) in the document mentioned in the next section (arXiv:math/0007024), has an example using the long unicode arrow `U+27F6`, while the one from tungsteno used the standard `U+2192`:
<img height="40" src="https://user-images.githubusercontent.com/348975/141471106-e9273e60-cbbd-4abb-a6dd-e67f11fd6572.png">

---

### n-ary subset-of-or-equal
The first display formula in [this latexml-converted arXiv:math/0007024 article](https://ar5iv.org/html/math/0007024):
<img height="40" src="https://user-images.githubusercontent.com/348975/141470514-b6685a56-2484-4812-b15b-a51f0f435cd8.png">

* Each operator is a simple element `<mo>⊆</mo>`, 
* Since the arguments are embellished, each is wrapped in a `<msubsup>`, so grouping in presentation follows content
* latexml succeeded parsing the structure in this specific case, so the n-ary operation is wrapped in its own `mrow`, and has clear scope. For instance, the trailing comma is *outside* the mrow.

---

### n-ary (argument) lists

[...]

Issue #252, opened 2021-11-09
Unifying intent


davidcarlisle:
At the call on 2021-11-04 two intent proposals were proposed (further discussion expected this week) This issue is the start at trying to propose a unified syntax,

----

## Bruce

https://mathml-refresh.github.io/discussion-papers/semantics-mini

```
intent   := literal | selector | intent '(' intent [ ',' intent ]* ')'
literal  := [letters|digits|_|-]+
selector := argref | argpath
argref   := '$' NCName
argpath  := '$' [digit]+ [ '/' [digit]+ ]*
```

## Sam

https://samdooley.github.io/mathml-docs/intent2cmml/intent.html

```
intent    :=  {explicit} | {implicit} | {reference} | {name} | {literal} | {special}
explicit  :=  {value} '(' ( {intent} [ ',' {intent} ]* )? ')'
implicit  :=  {value} '@' | '@' {value} | '@'
reference := '$' {name} | '$' {number}
literal   :=  '#' {name} | '#'? {number}
special   :=  '!' {name} | '!' | '/' {name} | '/'
value     :=  {reference} | {literal} | {name}
name      :=  [A-Za-z_][A-Za-z0-9_.]*
number    :=  [0-9.][A-Za-z0-9_.]*
```

---

I suspect we don't need `!`  even though it may help forcing some
specific content mathml forms such as `<apply><plus/>` rather than
`<apply><csymbol>+</csymbol>` we can probably arrange the defaults "do
the right thing" not dont need that,.

Similarly the literal prefix `#` is not I think needed as unprefixed
names and numbers can serve this purpose.  `<mi intent="3">iii</mi>`
in Bruces's proposal seems equivalent to `<mi intent="#3">iii</mi>` in
Sam's

The `intent=/` for empty is (normally) covered by having an intent on the
parent that doesn't select the element. I think remaining cases could
be covered by `intent=""`

Bruce confirmed in the call that the omission of a syntax for implicit
list of arguments was intentional but  I think that `(+ 1 2 3 4 5 6)`
becomes unweildy if we don't have this and even if we could make the
default intent of
`<mrow>... <mo>+</mo> ... <mo>+</mo> ... <mo>+</mo> ... <mo>+</mo>
... <mo>+</mo> ...`
be this, we can't write down the default value explicitly if there is
no syntax for narry application.

However I'm not convinced by the suggestion to allow @ on its own or
suffix or prefix position, it seems sound but is very hard to
[...]

PR #302, opened 2022-03-28
more words on intent

davidcarlisle:
This adds some more words to the stub `intent` section, I also tweaked the grammar (again) as the version there was I think too lax allowing named references within `/` steps

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -356,8 +356,7 @@ as an embellished symbol would be distinguished as follows:</p>
              &lt;/mfenced&gt;
            &lt;/mrow&gt;
          &lt;/mrow&gt;
-         &lt;annotation-xml cd="mathmlkeys" name="contentequiv"
-      encoding="MathML-Content"&gt;
+         &lt;annotation-xml cd="mathmlkeys" name="contentequiv" encoding="MathML-Content"&gt;
            &lt;apply&gt;
              &lt;ci&gt;a&lt;/ci&gt;
              &lt;apply&gt;&lt;plus/&gt;&lt;ci&gt;x&lt;/ci&gt;&lt;cn&gt;5&lt;/cn&gt;&lt;/apply&gt;
@@ -471,7 +470,7 @@ as an embellished symbol would be distinguished as follows:</p>
    <section>
     <h5>Attributes</h5>
 
-    <table border="1" class="attributes">
+    <table class="data attributes">
 
      <thead>
 
@@ -538,7 +537,7 @@ as an embellished symbol would be distinguished as follows:</p>
    <section>
     <h5>Attributes</h5>
 
-    <table border="1" class="attributes">
+    <table class="data attributes">
 
      <thead>
 
@@ -654,7 +653,7 @@ as an embellished symbol would be distinguished as follows:</p>
    <section>
     <h5>Attributes</h5>
 
-    <table border="1" class="attributes">
+    <table class="data attributes">
 
      <thead>
 
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -21,7 +21,7 @@
 
  <section>
   <h3 id="mixing_intent">The <code class="attribute">intent</code> attribute</h3>
-  <p>As decribed in <a href="#presm"></a> presentation <span class="ednote">Content? See <a href="https://github.com/w3c/mathml/issues/289">Issu 289</a>.</span>
+  <p>As decribed in <a href="#presm"></a> presentation <span class="ednote">Content? See <a href="https://github.com/w3c/mathml/issues/289">Issue 289</a>.</span>
   MathML elements
   (including [[MathML-Core]] allow attributes <code
   class="attribute">intent</code> and <code
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -1,11 +1,11 @@
 <section data-max-toc="3">
  <h2 id="mixing">Annotating MathML</h2>
- <div class="issue" data-number="281"></div>
+<!-- <div class="issue" data-number="281"></div>-->
  <div class="issue" data-number="252"></div>
  <div class="issue" data-number="289"></div>
 
  
- <p>Semantic annotation refers to markup added to
+ <p>Annotation refers to markup added to
  make associations between alternate representations of an
  expression, and for associating semantic properties and other
  attributions with a MathML expression.</p>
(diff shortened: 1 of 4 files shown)


src/mixing.html
```diff
@@ -99,9 +124,9 @@ <h3 id="mixing_msup_notation">Same notation, different meanings</h3>
 &lt;/msup>
 </pre>
 
+ 
  </section>
  </section>
-
  
  <section>
   <h3 id="mixing_semantic_annotations">Annotation Elements</h3>
```
dginev: Optional nitpick: The id and refs to it here may also switch from `mixing_semantic_annotations` to `mixing_annotations`.

src/mixing.html
```diff
@@ -15,53 +15,78 @@ <h2 id="mixing">Annotating MathML</h2>
...
-literal      := [letters|digits|_|-]*
-selector     := ( argref | numref ) [ '/' selector ]?
-argref       := '$' NCName
-numref       := '$' [digit]+
-wildcard     := '@'
+intent   := literal | selector | intent '(' wildcard ')' | intent '(' intent [ ',' intent ]* ')'
+literal  := ( letters | digits | '.' | '_' | '-' )*
+selector := argref | numref
+argref   := '$' NCName
+numref   := '$' [digit]+ ( '/' numref )?
```
dginev, line 101: First, isn't `numref` actually a `pathref` given the slashes?

For the record I'm still not fond of including numbers or paths given that `argref` is more human-friendly and covers an equivalent class of expressions. Having two separate referencing schemes is a bit overkill for a simple mini-language.

I probably don't have enough group support to stop including both, but if I did - I would.

src/mixing.html
```diff
@@ -15,53 +15,78 @@ <h2 id="mixing">Annotating MathML</h2>
...
 
    <pre>
-intent := literal | selector | intent '(' wildcard ')' | intent '(' intent [ ',' intent ]* ')'
-literal      := [letters|digits|_|-]*
-selector     := ( argref | numref ) [ '/' selector ]?
-argref       := '$' NCName
-numref       := '$' [digit]+
-wildcard     := '@'
+intent   := literal | selector | intent '(' wildcard ')' | intent '(' intent [ ',' intent ]* ')'
+literal  := ( letters | digits | '.' | '_' | '-' )*
```
dginev, line 98: How about letters, digits and dash? Why do we need dot and underscore?

src/mixing.html
```diff
@@ -15,53 +15,78 @@ <h2 id="mixing">Annotating MathML</h2>
...
+selector := argref | numref
+argref   := '$' NCName
+numref   := '$' [digit]+ ( '/' numref )?
+wildcard := '*'
```
dginev, line 102: I understand this is meant to help with the n-ary infix cases, but *does it* help with the n-ary infix cases? I'm not so sure, given that it doesn't mean "all children", but "all children that the author meant were arguments".

Does it matter if one does:
```xml
<mrow intent="cross-product(*)">
  <mo>(</mo>
  <mi>A</mi>
  <mo>×</mi>
  <mi>B</mi>
  <mo>×</mi>
  <mi>C</mi>
  </mo>)</mo>
</mrow>
```

vs
```xml
<mo>(</mo>
<mrow intent="cross-product(*)">
  <mi>A</mi>
  <mo>×</mi>
  <mi>B</mi>
  <mo>×</mi>
  <mi>C</mi>
</mrow>
<mo>)</mo>
```

or is it *only* going to work on arg-annotated trees?
```xml
<mrow intent="cross-product(*)">
  <mo>(</mo>
  <mi arg="A">A</mi>
  <mo>×</mi>
  <mi arg="B">B</mi>
  <mo>×</mi>
  <mi arg="C">C</mi>
  </mo>)</mo>
</mrow>
```

I am unconvinced it is really needed at this point in time.

src/mixing.html
```diff
@@ -15,53 +15,78 @@ <h2 id="mixing">Annotating MathML</h2>
...
+  of</p>
+  <ul>
+  <li><em>references</em> prefixed with <code>$</code>. A numeric
+  reference such as <code>$3/$1</code> references the [=intent=] of
+  the first child of the 3rd child of the element. An argumnt
+  reference such as <code>$name</code> refereces a descendent element
+  that has an attriute <code>arg="name"</code>. unlike <code class="attribute">id</code>
+  attributes, <code class="attribute">arg</code> do not have to be
+  unique within a document, when searching for a matching element the
+  search should only consider descendns and not consider descendents
```
dginev: `descendns` typo.

Maybe: " should only consider descendants, while stopping early at any nodes that have a set intent or arg attributes, without descending in them."

src/mixing.html
```diff
@@ -15,53 +15,78 @@ <h2 id="mixing">Annotating MathML</h2>
...
+  of any element that has an <code class="attribute">intent</code> or
+  <code class="attribute">arg</code> attribute.</li>
+  <li><em>literal numbers</em> such as <code>2.5</code> denote
+  themselves. <span class="ednote">0.25e1?</span></li>
```
dginev, line 53: Are you trying to get AT to pronounce "zero point twenty-five e 1"? What kind of `<mn>` (or other element) needs this type of annotation?

src/mixing.html
```diff
@@ -15,53 +15,78 @@ <h2 id="mixing">Annotating MathML</h2>
...
+  href="https://docs.google.com/spreadsheets/d/1EsWou1K5nxBdLPvQapdoA9h-s8lg_qjn8fJH64g9izQ/edit#gid=1358098730">Google
+  Sheets</a>. Note the list of [=intent=] identifiers is
+  <em>Open</em>: level 1 is a curated list maintained by the Math Working
+  Group and should be implemented by any implemenation of MathML
+  intent, so that suitable audio renderings of the level 1 terms will
+  be generated. Terms in other levels may be contributed; the
+  dictionary may give speech hints, but a system may also fall back on
+  reading the identifier name as given.</li>
+  <li><em>the wildcard</em>, <code>*</code> denotes all the children
+  <span class="ednote">just mi and mn children?</span> of the element so
```
dginev: certainly not just mi and mn, since also msub, msup, msubsup, mover, munder, munderover, mrow, msqrt, mfrac, mroot, mtable, mstyle, mpadded, ... can contain valid arguments even to simple numeric functions

src/mixing.html
```diff
@@ -15,53 +15,78 @@ <h2 id="mixing">Annotating MathML</h2>
...
-
-<p>On the other hand, mathematics encompasses an endless set of concepts, arguing that the set of meanings must be open-ended. We propose that there should be a small set of recommended meaning keywords whose translation can be specialized, while allowing any value for the meaning (an implementation is free to recognize more keywords). In the case where a meaning is not in a dictionary, support for multiple languages may be weak and the application of such a literal would be generic as “the [meaning] of [arg1], [arg2] and [argn] …”. While this result may be less than optimal, it is still functional.</p>
   </section>
 
+
  
  <section>
-  <h3 id="mixing_msup_notation">Same notation, different meanings</h3>
+  <h3 id="mixing_intent_examplesn">Intent Examples</h3>
```
dginev, line 110: is `examplesn` a typo? maybe not, but thought I'd ask.

dginev (commented):
This is a very exciting evolution of the intent text and details, thank you David!

Unsurprisingly I have some reservations and comments on the fine-print details, for your consideration.

But I am really happy about the way the word "semantic" is receding to places where it has technical bearing, and we're using it less in the colloquial sense. Those parts read much better to me.

src/mixing.html
davidcarlisle: pathref yes perhaps. It's only a name.

Actually that was a very late change just before I pushed as I didn't want to have to document what `$foo/$3/$bar` (allowed in the existing gh-pages draft) means so I decided to disallow it.

I'm also not totally convinced by the need to have `/` (or `*`)  but I'd rather leave them in for now until we have some words and/or implementation around default values for `intent` If we end up not needing this to describe `intent` then we can probably drop it.  

Using `$argref` can make the same references but  requires modifying the child elements which may not be so convenient. That is, when generating a non-fraction mfrac you may be able to generate `mfrac intent="fooobar($1,$2)"` then forget about it, if you generate `mfrac intent="fooobar($a,$b)"` then your generator needs to recursively "remember" to add `arg="a"` and `arg="b"` in the right places. So it  probably ends up being simpler to specify default intents using a path. If you are adding intents by hand to an existing tree, using argref is simpler I agree.

src/mixing.html
davidcarlisle: underscore was in one (all? some?) of the intent grammars proposd in the mathml-refresh it just got copied in here, I have no particular argument in favour of it. I added `.` in this PR as I think if we allow `intent="2"` we have to allow `intent="2.5"`

src/mixing.html
davidcarlisle: Me too:-) I think that's the most "at risk" part of the grammar, it is temping but specifying what exactly it picks up is problematical. That's really why I am pushing to get a grammar in the spec, at least then we all look at exactly the same version and if it's wrong we can change it.

I think that it probably has to mean "any mi or mn child" so all three of  the cases you show work, but if you add any redundant mrows it doesn't match. But maybe that doesn't work well enough but if we end up with some complicated algorithm to say what it should select then it should go.  Again I'd like to see some specifications of default intent with and without this.

src/mixing.html
davidcarlisle: It (again) depends what we want to say about defaults, if the default intent of an mn is its text content then `<mn>2.5</mn>` should (perhaps) be equivalent to `<mn intent="2.5">2.5</mn>` which means that should be allowed.

 Also if in general we want to allow intent to be "moved up the tree" to give less fine grained control but more control over how the entire expression is spoken then you need to allow

```
<mrow intent="times(2.5,$x)">
  <mn>2.5</mn>
   <mo>*</mo>
  <mi>x</mi>
</mrow>
```

src/mixing.html
davidcarlisle: yes I know:(  As I say above I think the specification of what this picks up has to fit in to a sentence, otherwise it has to go, I really don't want to have a whole section on a matching algorithm just on how the wildcard works. It's intended to make things simpler.  Open to suggestions........

src/mixing.html
davidcarlisle: sorry, it's statistically reasonable to assume typos if I typed it.

src/mixing.html
davidcarlisle: better wording than mine

src/mixing.html
davidcarlisle: yes, alhough I want it to be the `<semantics` annoaion rather than `intent=` I'll make it `mixing_annotation_elments` (actually we should probably change the `mixing_` prefix as well but that would chang refrences over he document (and the notes document) so leaving that for now to keep the diff here localised.

src/mixing.html
davidcarlisle: perhaps the default should be "every child  except (possibly embellished) operators".  Matching embellished operators is a bit non-trivial but at least it is not new text: it's an already defined concept.

src/mixing.html
dginev: > If we end up not needing this to describe intent then we can probably drop it.

For the record, *I* currently don't need this to describe intent, and would be quite OK with dropping it. I'm really not fond of preemptively adding untested bits to the grammar, *given that* we have tested a wide range of expression with a more constrained syntax.

Just as one example of unintended side-effects, the grammar above now allows the following as a valid intent value:
```
.(.. , ..., , , ,)
```

To my taste, the literal rule is along the lines of:
```
literal  := letter+ ('-' letter+)*
```
focusing on the encyclopedic intent setup, where we have ample examples for the concrete needs of AT. The numeric annotations strike me as somehow off-topic until we get some examples where they appear to be a must-have.

src/mixing.html
dginev: I'm not sure that's a *need*. The presentation tree is primary, and it can contain more information than the set allowed in the intent attribute itself. If we specify that

> the default behavior on leaf "token elements" is to read out their text content

then the English stating that suffices. 

If, instead, the aim was to allow rewriting arbitrary presentation MathML trees *into* intent values, then the "literal" grammar rule has to extend to all of Unicode, not just letters or numbers. Especially when we remember that `<mtext>` is also in the mix.

src/mixing.html
dginev: Indeed, `"every child except (possibly embellished) operators"` is better.

Not having the wildcard at all would be even nicer, given that n-ary infix operations can simply be left "as-is", without *any* intent annotation. Still waiting for counter-examples in #253 on that point.

src/mixing.html
davidcarlisle: > > If we end up not needing this to describe intent then we can probably drop it.
> 
> For the record, _I_ currently don't need this to describe intent, and would be quite OK with dropping it. I'm really not fond of preemptively adding untested bits to the grammar, _given that_ we have tested a wide range of expression with a more constrained syntax.

well that's rather the point, you, Sam, Bruce, Neil, then I all had different syntaxes in intent suggestions.
We need to converge on a single syntax.

> 
> Just as one example of unintended side-effects, the grammar above now allows the following as a valid intent value:
> 
> ```
> .(.. , ..., , , ,)
> ```

actually that was an _intended_  feature, the grammar I originally had in the unified intent issue did not allow empty arguments in the intent function syntax, just allowed a complely empty intent. @brucemiller  commented that his grammar had allowed, and used empty argument slots. The grammar only needs to constrain so far, it will naturally allow some non sensical expressions which need to be handled at a higher level. We could change some `*` to `+` if empty slots turn out  not to be needed, but we'll see,. It would of course be possible to have a grammar that allowed empty argument slots but forced a non empty identifier in the function name, but if we want to allow empty anyway keping the grammar simple seemed preferable rather than forcing special non-empty/empty productions just for this. So I allowed empty everywhere in this draft.

> 
> To my taste, the literal rule is along the lines of:
> 
> ```
> literal  := letter+ ('-' letter+)*
> ```
> 

see above...

> focusing on the encyclopedic intent setup, where we have ample examples for the concrete needs of AT. The numeric annotations strike me as somehow off-topic until we get some examples where they appear to be a must-have.

we had a general aim that people should be able to do these attributions at different levels of granularity. If you want to go down to the leaves and have  `<mn>1</mn>` then some designs would advocate always having an explicit possibility for stating the default and that would seem to be `<mn intent="2">2</mn>`  More usefully if you want to annotate larger expressions without annotating the whole tree you might want to end up with  `<math intent="plus(power(cos(x),2),power(sin(x),2))"`  (by the way if we are aiming for accesibility rather than computation, why don't we allow infix functions in intent?, I want to write  `intent="power(sin(x),2)+power(cos(x),2) = 1" but in any case these seem to use numberic literals in intent.

The semanitics-mini paper had `power(x,2)` in is opening examples, which used 2....

dginev:
I just had the most productive idea:

Could you wrap any grammatical details around `intent` inside an `ednote` span, making it clear this is still to-be-decided and merge the PR as-is?

With that in place, it's good to merge for me.
We'll have to restart the grammar consensus talks at some point, but this is likely the wrong place.

src/mixing.html
dginev: >  it will naturally allow some non sensical expressions which need to be handled at a higher level.

Well, the nonsensical uses of dot and underscore are very easy to prevent in the grammar itself, similarly to how i guarded the dash in my comment above. The higher-level problem is that if we add constructs without double and triple-checking their utility, we end up with a messier outcome that has a greater learning curve for less clarity.

An empty argument pattern is just ` '(' ')'` which can be allowed standalone, if need be. Though here again I'm interested in what AT purpose that serves, since it looks suspiciously close to a programming function call, rather than an AT aid.

> an explicit possibility for stating the default

My worry is that this line of reasoning eventually implies full Unicode support in literals, not just numbers. See the start of [arXiv:1810.00769](https://ar5iv.labs.arxiv.org/html/1810.00769) for one example:
```xml
<mi intent="♣">♣</mi>
<mi intent="♡">♡</mi>
```
People make use of a lot of characters beyond letters and digits in their MathML token elements, especially `mi` and `mo` nodes, as I'm sure you know. Stating the default intent to be "as written", requires us to allow hundreds, if not thousands, of characters in the `literal` rule. On top of that, having a grasp of compound objects, such as numbers in scientific notation, requires additionally growing the grammar if we wanted to validate them.

From where I'm standing, none of this extra boilerplate is needed. If one had to, they could use an arg reference in the complex expressions, as in `power(speed-of-light, $2)`. And, most productively, we can completely avoid annotating nodes that follow the specified defaults. `<mn>2</mn>` should never need a redundant `intent="2"`.

As another alternative, XPath has the single `text()` selector for text content. Allowing that one special keyword exception appears to me as more manageable than opening the door to all of Unicode, and/or growing the grammar complexity.

---

The challenging example I initially was fond of was `second-derivative`, where "the second derivative of f" could be written:
```xml
<msup>
  <mi>f</mi>
  <mo>″</mo>
</msup>
```
One way to mark that up would be `derivative(f,2)`, where the `2` has no literal equivalent in the MathML, and would require numbers added to the `literal` rule. 

But there is another way - using a [second-derivative](https://en.wikipedia.org/wiki/Second_derivative) literal, as in `second-derivative(f)`. Having an encyclopedic page for that string naturally shifts my own preference towards using that keyword, with no argument beyond the f (or a `$ref` to f), instead of an open-ended numeric argument.

brucemiller: If we want to have an open-ended components (ie. not only keywords specified in some to-be-determined dictionary), then we will have to allow (most of) Unicode.  I would have been inclined to allow literals to include *all* characters not explicitly given some specific grammar here (parens, comma, $), but that severely limits any future expansion.  Slightly safer approach would reserve a set of punctuating chars for potential future use.

Explicitly marking arbitrary literals using `text()`, or quotes (embedded, argh!) may be safer.  But that opens up more potential for user error. And also seems to introduce a distinction between open ended text and keywords (That could be argued as either good or bad).

davidcarlisle: > An empty argument pattern is just '(' ')' 


yes but the example needed empty arguments between the commas, no just an empy argument list, so `foo(a,b, , )` was the intended syntax in Bruce's proposal if I undersood it right.

> But there is another way - using a [second-derivative](https://en.wikipedia.org/wiki/Second_derivative) literal

yes you can always do that but it doesn't really scale, for derivatives you might get away with two or three fixed idntifiers but other things will look increasingly strained if you do that.

As for unicode, we may not end up allowing card suits but I can't see us only allowing leters A-Z we'd have to allow all unicode characters with class letter or XML NCName or some such. I doubt it would get past the CR review if we only allow latin idenifiers.

That said we haven't even reached first public working draft stage yet so the whole spec is tentative, so I've no objection to sticking a big green ednote box by the grammar saying the grammar isn't decided, I thought that was already implied by having the open issue #252 taking up a couple of screenfuls at the top of that section, but I'll add an ednote to the grammar:-)

davidcarlisle:
@dginev is that clear enough?

dginev:
@davidcarlisle it's good enough to merge for now - thanks!

src/mixing.html
davidcarlisle: @brucemiller  yes I agree that we will need most of unicode, taking NCName might be a coherent choice as that avoids most punctuation and has the advantage of being someone else's arbitrary choice. Otherwise I was assuming that `letter` in the grammar meant `[:letter:]` ie whatever Unicode says is a letter, not `[a-z]`

dginev: My working assumption for letters has been exactly "whatever Unicode says is a letter", since a main hope for the literals is to represent "concept names". And names are comprised of alphabetic letters, in whichever language an author is happening to write them down.

But I also never wanted to be providing tooling for literal content *other than* concept names. A very minimal and constrained initial scope of intent.

NSoiffer:
Seems like plenty of comments so far. I'm merging for the meeting and people can always comment/make changes.

PR #305, opened 2022-03-31
Global changes:

NSoiffer:
1. Went through and adding folding to all the examples where it seems appropriate
2. Marked all the elements and attrs that are in/out of core appropriately.

NOTE: there seems to be some style bug where class="coreyes" in the tables of attribute values doesn't render anything.

Started going through the chapter pulling out informative sections and doing some rewrites in their place. Still much more to go!

I introduced a glossary reference to the notes because I didn't think it made sense to do full linking everytime I said "see the notes for usage information". This will geneate a respec error.

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -2180,7 +2180,7 @@
       <tr>
        <td colspan="3" class="attdesc">
 	Specifies whether the operator should be kept symmetric around the math
-	<a class="intref" href="#dt-axis">axis</a> when stretchy.
+	[=axis=] when stretchy.
 	Note this property only applies to vertically stretched symbols.
 	See <a href="#presm_op_stretch"></a>.
        </td>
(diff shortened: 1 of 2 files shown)


src/presentation-markup.html
```diff
@@ -2115,7 +2115,7 @@ <h6 id="presm_mo_dict_attrs">Dictionary-based attributes</h6>
        <td rowspan="2" class="attname">separator</td>
        <td>"true" | "false"</td>
        <td><em>set by dictionary</em> (false)</td>
-       <td><span class="coreyes"></span></td>
+       <td><span class="coreno"></span></td>
```
davidcarlisle, line 2118: This is the one coreyes/no I changed (I don't think core has separator)

davidcarlisle (approved):
Looks good, I added MathML-Notes to the local biblio so links now work, and made the coreyes span into links to core, also changed axis to use respec `[=axis=]` markup (as links to the old glossary entry broken as glossary gone.

still a few respec warnings about definitions but mosly relating to length setup in chapter 2  that is pending on deciding about mpadded length syntax mosly. only warnings now, all respec errors gone, so we can ignore for a while.

PR #308, opened 2022-03-31
Move p2s transformation to a new appendix

samdooley:


src/transform2strict.html
```diff
@@ -0,0 +1,2704 @@
+
+  <section>
+   <h3 id="contm_p2s">The Strict Content MathML Transformation</h3>
```
davidcarlisle, line 3: it doesn't actually make any difference as respec ignores which h? element is used and just goes off section nesting, but it might be less confusing to lift all the sections one level as this source h3 will generate a h2 now.
But that can be done  later, I'll merge as is now.

davidcarlisle (approved):

PR #310, opened 2022-04-01
Removed deprecated declare, fn, reln

samdooley:


davidcarlisle (approved):
yes I'll add something to the changes appendix to match

PR #311, opened 2022-04-01
Davidc/mfenced

davidcarlisle:
This PR removes `<mfenced>` from the sample presentations in chapter 4, making them potentially viewable with mathml core


NSoiffer:
As I mentioned in some other issue, we will have to face the problem of displaying non-mathml core code for other cases. But I'm ok with deemphasizing `mfenced`.

PR #312, opened 2022-04-02
Davidc/inline mathml

davidcarlisle:
This PR adds inline mathml after core mathml examples, styled with a thick left border to match the current w3c /respec styling.
The rendered div is removed on save so would no appear in official WD or REC versions but appears in editor's drafts with live respec processing.

I think it's a useful test that the examples actually are examples, when adding `&lt;` quoted matthml inside an hml div it's easy to get something wrong, and this provides immediate feedback.

image


davidcarlisle:
converted to draft. 

locally I had rebased this branch and it showed a clean diff with just a single javascript function added in index.html and no changes elsewhere, bu this PR shows commits back to the dawn of time and merge conflicts.  I'll sort out and resubmit.

davidcarlisle:
closing this, a new PR #313 from a clean branch gets to the same endpoint.

PR #313, opened 2022-04-02
inline mathml rendering in respec drafts

davidcarlisle:
This (fixed) PR adds inline mathml after core mathml examples, styled with a thick left border to match the current w3c /respec styling.
The rendered `<math>`  is removed on save so would not appear in official WD or REC versions but appears in editor's drafts with live respec processing.

I think it's a useful test that the examples actually are examples, when adding `&lt;` quoted mathml inside an html div it's easy to get something wrong, and this provides immediate feedback.

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -75,7 +75,18 @@
     }
     function resortable () {
         sorttable.makeSortable(document.getElementById('oper-dict-table'));
-      }
+    }
+    function showcore (){
+	var mmllist=document.querySelectorAll("div.example.mathml");
+	for(var i=0;i!=mmllist.length;i++){
+	    var m=document.createElementNS("http://www.w3.org/1998/Math/MathML","math");;
+	    m.setAttribute("display","block");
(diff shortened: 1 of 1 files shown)


davidcarlisle:
Actually currently this shows rendering for all mathml, even content mathml which can't work without some polyfill which I don't think we want here. Currently we are using `<div class="example mathml">` for all mathml examples but it would be easy to adjust the css as the "sample presentation" examples are very stylised (having originally been auto-generated)  so a simple global edit. We can do that separately to this PR (and should probably do anyway, evn if we don't merge this PR)

NSoiffer:
I'm doing the merge because I think this is a good idea.

Because examples are full MathML, they don't necessarily have renderings in browsers that only support core. Hence, I think polyfills will be needed. One approach is that each example potentially mention a polyfill needed for rendering it (if not 100% core), or maybe the whole spec just bites the bullet and loads all the polyfills.

PR #315, opened 2022-04-04
clipboard examples

davidcarlisle:
This PR deletes section 6.3.4  (clipboard examples) and makes no other changes.

The text has been added to the notes-on-mathml to the previously empty notes file for chapter 6

https://w3c.github.io/mathml-docs/notes-on-mathml/#clipboard-examples

(I added it to the notes already, just to avoid losing text, but that can be reverted if this PR is not merged and the examples stay in the main spec)

The formating of the tables in the notes is approximately markdown tables extension syntax, which doesn't seem currently supported by respec, but that can be addressed later, it is aligned so readable as a `<pre>` block.


polx:
Just for me to understand: DId we discuss removing this or is there a basic reason?

davidcarlisle:
@polx  not other than the "general plan" as outlined in the status of this document of moving non normative text out of the spec to the notes. I was just flicking through the whole spec looking for chunks of text to populate the notes document and his seemed to be the largest chunk of (presumably non normative) examples.

If you'd rather consider this in sequence when editing the chapter that's fine, can close this with no action, it's only a "kill region" in emacs:-)

The  "status of this document"  section (text from @NSoiffer  I guess) says

> Chapters 6 and 7 contain very useful informative information on interacting with HTML, XML, etc., along with usage suggestions and warnings about specific characters. Because these are informative, these two chapters will be removed from the MathML Full spec and move to the notes.

so moving a least one section of chaper 6 seemed to fit that plan. But doing it in random order (or at least order of size of examples) may not be best.

davidcarlisle:
@polx  I plan to close this with no action (otherwise would have to fix the merge conflicts have have arisen) Happy for you to do whatever seems right when revising this chapter, this was just a suggestion. The examples (converted to markdown) are in the notes draft already. I will leave them there, but similarly they can stay or be removed when the final form is clearer.

PR #323, opened 2022-04-21
A little more work on presentation. Removed a few things and moved th…

NSoiffer:
…e linebreaking examples to be part of the discussion on linebreaking (was sitting by itself).

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -1,1586 +1,1016 @@
 <section>
- <h2 id="presm">Presentation Markup</h2>
-
- <div class="issue" data-number="271"></div>
- <div class="issue" data-number="272"></div>
-
- <section>
-  <h3 id="presm_intro">Introduction</h3>
-
-  <p>This chapter specifies the <span>&#x201c;presentation&#x201d;</span> elements of
-  MathML, which can be used to describe the layout structure of mathematical
-  notation.
-  </p><p>
-  Most of Presentation Markup is included in [[Mathml-Core]].
-  That specification should be consulted to for the precise details of displaying the elements and attributes
-  that are part of core when displayed in web browsers.
-  Outside of web browsers, MathML presentation elements only suggest (i.e. do not require)
-    specific ways of rendering in order to allow for medium-dependent
-    rendering and for individual preferences of style.
-    Non browser-based renderers are free to use their own layout rules as long as the
-    renderings are intelligible.</p>
-  <p>The names used for presentation elements are suggestive of their visual layout.
-    However, mathematical notation has a long history of being reused as new concepts are developed.
-    Because of this, an element such as <code class="element">mfrac</code> maybe not actually be a fraction and the
-    <code class="attribute">intent</code> attribute should be used to provide information for auditory renderings.
-  </p><p>
-  This chapter describes all of the presentation elements and attributes of MathML along with examples that might clarify usage.
-  </p>
-
-
+  <h2 id="presm">Presentation Markup</h2>
+ 
+  <div class="issue" data-number="271"></div>
+  <div class="issue" data-number="272"></div>
+ 
   <section>
-   <h4>Presentation MathML Structure</h4>
-   <p>The presentation elements are meant to express the syntactic
-   structure of mathematical notation in much the same way as titles, sections,
-   and paragraphs capture the higher-level syntactic structure of a
-   textual document. Because of this, a single row of identifiers and operators
-   will often be represented by multiple nested <code class="element">mrow</code> elements rather than
-   a single <code class="element">mrow</code>.  For example, <span>&#x201c;<i class="var">x</i> + <i class="var">a</i> /
-   <i class="var">b</i>&#x201d;</span> typically is represented as:
-
-   <div class="example mathml">
-    <pre>
-      &lt;mrow&gt;
-        &lt;mi&gt; x &lt;/mi&gt;
-        &lt;mo&gt; + &lt;/mo&gt;
-        &lt;mrow&gt;
-          &lt;mi&gt; a &lt;/mi&gt;
-          &lt;mo&gt; / &lt;/mo&gt;
-          &lt;mi&gt; b &lt;/mi&gt;
-        &lt;/mrow&gt;
-      &lt;/mrow&gt;
-    </pre>
-   </div>
+   <h3 id="presm_intro">Introduction</h3>
+ 
+   <p>This chapter specifies the <span>&#x201c;presentation&#x201d;</span> elements of
+   MathML, which can be used to describe the layout structure of mathematical
+   notation.
+   </p><p>
+   Most of Presentation Markup is included in [[Mathml-Core]].
+   That specification should be consulted to for the precise details of displaying the elements and attributes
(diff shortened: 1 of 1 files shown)


davidcarlisle:
@NSoiffer  I resolved the merge conflicts locally (in favour of your branch in all cases) and pushed to gh-pages but you might want to check the outcome....

dginev:
General comment on merging huge diffs: someone needs to take a week off to really make sure the changes are as an intended.

A `git rebase main` on the side of the PR author (Neil here) may have dramatically reduced the size of the changes instead. With nearly 10,000 lines changed no one can really know what happened in this PR.



If you want to have confidence you still know what is in the main branch, I suggest rewinding to its state before the PR (commit 7ede83b3e4f202c44a960a7f2da0aa96ef28866d ) and then simply cherry-picking Neil's initial commit into it via:
```
git checkout main
git reset --hard 7ede83b3e4f202c44a960a7f2da0aa96ef28866d
git cherry-pick 5a2e688403373e7cdda57b9f67fd7d0f2506fa6a
git push --force
```

Rewinding and avoiding the merge step, and the huge diffs, should also make the new #324 disappear.

davidcarlisle:
@dginev  I did actually compare the diffs to be satisfied to merge, There were not really 10000 lines (of the diff presented),  most of it is white space, which I normalised locally while checking, there were then just 25 merge chunks to inspect and as I commented I resolved in favour of the feature branch in all cases, except one spurious end tag so the end result of the PR merge is that it is exactly as in @NSoiffer's branch except for the diff 


```
$ git diff presentation -- src/presentation-markup.html 
diff --git a/src/presentation-markup.html b/src/presentation-markup.html
index d471686..c2cfb17 100644
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -3136,7 +3136,7 @@
     &lt;/mrow&gt;
  </pre>
  </div>
- </p>
+
  </section>
  
  <section>
@@ -9722,4 +9722,4 @@
  </section>
  </section>
  
- 
\ No newline at end of file
+ 
```

All the real text edits in chapter 3 are from Neil and so there would not normally be any conflicts and we just decide to merge or not as a block if we approve the intent of the edit.  The conflicts here are with a technical global structural edit I had done earlier relating to display of examples. It's unfortunate that this wasn't pulled to the presentation branch before the last round of edits, but sorting it out isn't something we need to take a week off to discuss.

I considered blocking the PR due to the respec errors due to broken links, but even here it seems better to merge, let people see the rendered version on the main site, and open a new issue (#324) to fix up the spec post merge,

> Rewinding and avoiding the merge step, and the huge diffs, should also make the new https://github.com/w3c/mathml/issues/324 disappear.

I assumed it was intended to remove that section. If that wasn't the case we can always back out the PR, as you note.

dginev:
I hope I am clearly marking the difference that currently only @davidcarlisle knows if the PR was merged correctly, as opposed to the PR itself having an informative diff that anyone else can review.

To clarify, I didn't mean taking a week off to "discuss", I meant it as literally taking a week to pour over the 10k diff, line by line, to make sure there isn't any issue sneaking in from the git merge. This is also where a `rebase` from Neil, instead of the merge, can lead to much clearer diffs in the PR "Files changed" tab.

Maybe it was indeed mostly a difference in whitespace, and maybe you have a great local process to normalize these quickly and feel confident no change slipped by. But I have no way to know if that's true without figuring out a similar local setup and reproducing these steps on my own machine. The transparency of using PRs gets lost, in my eyes.

davidcarlisle:
@dginev I wrote a longer reply but deleted it:-)

I think we are talking cross purposes. I have tried to explain my approach to resolving the merge conflict in the PR (which wasn't so big really)  whereas I guess your main issue is with the merge done to the feature branch before making the PR (which was a bit big)  and you would have commented about that even if there was no conflict in the PR and I had been able to merge the PR simply in the website GUI.  I don't disagree with the general point, but I don't think any harm done in practice (or rather trying to force push a different history after the event has the danger of doing more harm than good in terms of transparency)

We have large amounts of text still to write (intent.....) it really don't agree  it would be useful to spend a week reviewing  line by line diffs on a change such as this which as it says in the title is a minor change,  deleting a few paragraphs and moving some. Review the resulting document, the line by line diff is just what it is, a necessary artifact of mechanical changes.

dginev:
Apparently we are not reaching each other with the content behind our comments. Let me try again:

> really don't agree it would be useful to spend a week reviewing line by line diffs on a change

Exactly! That is precisely my point - these large artifact diffs need to be avoided, or github's PR interface and git's history become unusable. Then we can't do any tracing back through the versions when a problem is discovered at some future date. And we can't catch it preemptively while the PR is still open.

Using `rebase` could avoid causing the artifacts for the PR author, while using `reset` and `cherry-pick` with a little `push --force` could restore the history to a usable state even after the merges have slipped in.

> but I don't think any harm done in practice

How would you (and worse: others) know if there was, without reviewing the 10,000 lines in question? We *hope* there is no harm done in practice. For example, months may pass before someone notices a small example is missing somewhere, if it were to slip through the merge conflict resolution.

> force push a different history after the event has the danger of doing more harm than good in terms of transparency

I believe having the record of this public discussion in the PR comments also avoids that issue.

NSoiffer:
I did a pull before making the changes, but in hindsight, I should have done
```
git rebase main
```
as @dginev suggested. My bad. Or maybe I should close the branch out after each commit and start a new one.

davidcarlisle:
@dginev 

> exactly! That is precisely my point - these large artifact diffs need to be avoided

I don't think that's an achievable aim. 

Look for example at the parallel discussions on re-arranging chaper 4.  I haven't yet made a PR but whether you look a  Sam's proposed re-arrangement or mine, there are some paragraphs describing the _intent_ of the change (in my case "merge each section of 4.4 into the corresponding subsection of 4.3") so we can discuss that but if we decide to do it a some point a line diff of
https://davidcarlisle.github.io/mathml/#contm2
against
https://w3c.github.io/mathml/#contm
will be made: that's hundreds of  paragraphs moved, or deleted, it will be a much bigger diff than this and viewing it as a line by line diff will be mostly pointless, or a least if you review the diff in that way you need some semi automated help such that the size of it isn't an issue.

But I can only make this comment as you over-stated your case. Large diffs can not be avoided, this particular one would have been smaller with a rebase but we both know that.

davidcarlisle:
@NSoiffer 

> I did a pull before making the changes, 

Yes that's how I read it.

>  Or maybe I should close the branch out after each commit and start a new one.

Possibly, but keeping a long running branch is OK as well but it does help if you re-sync it to gh-pages before doing anyhing major:-)

dginev:
> But I can only make this comment as you over-stated your case. Large diffs can not be avoided, this particular one would have been smaller with a rebase but we both know that.

My case was not overstated, rather misunderstood. I mentioned large *artifact* diffs, not *all* large diffs. Merges in particular. I think I can only repeat this so many times without sounding like a broken record. This is one of the pieces to "we are using git" as opposed to "we happen to be using git".

If the difference was a key part of the PR, of course it should be viewable - I've looked at Sam's PRs, and had no need to raise any points.

Issue #320, opened 2022-04-08
chapter 4 rearrangement sketch


davidcarlisle:
In #319 @samdooley  proposes a re-organisation of chapter 4 removing the syntax tables from each element.

As I comment in the PR I think this goes too far in removing information from the main section for each element.

I think the main section should have some formal statement that `<plus/>` is empty used with `<apply><plus/>` not a container used as `<plus><cn>1</cn><cn>2</cn></plus>`

Rather than just make negative comments I thought I'd sketch a different direction for condensing this material, although it's at a preliminary stage, and I am not generating a PR.

The basic idea is that currently each section such as the one for `<plus/>` has a table

```
Class                nary-arith
Attributes        CommonAtt, DefEncAtt
Content           Empty
Qualifiers        BvarQ,DomainQ
OM Symbols   [plus](https://openmath.org/cd/arith1#plus)
```

having such a table for every entry does get very verbose,  but acually only the first row is needed, once you know `<plus>` is of class `nary-arith`  you know it is empty and its allowed attributes, and all elements of  that class map to OM arith1 CD symbol of the same name as the the element..

So the basic idea of the fork is to compress the table to one row then put a "template" for all elements where the class is introduced in section 4.3.4

plus:  https://davidcarlisle.github.io/mathml/#contm_plus

conm-nary: https://davidcarlisle.github.io/mathml/#contm_nary

Currenly the same styling is used so the 1-row table looks a bit spacy...

----

This simplification breaks down a bit in cases  where the mapping to OM isn't 1-1 or where there are extra qualifiers eg `<degree/>` but I think with some careful wording that could be managed....  

---

This fork also shows two other unrelated simplifications

 * Content mathml examples are not inlined to be rendered
 * Presentation  MathMl sample renderings  _only_ have the inline mathml not the image (ie support for core is assumed)

(For ease of experimenting, in this sketch they are there just hidden with css)

But all in all 

https://davidcarlisle.github.io/mathml/#contm_plus

is a lot shorter than

https://w3c.github.io/mathml/#contm_plus

physikerwelt:
To me, Section 4.4 has a large overlap with the new intent markup. Maybe we should deprecate the entire set of pragmatic content in favor of the intent attribute. Otherwise, we have two ways of specifying somehow semiformal mathematical expression trees.
If we decide to do so, we can keep the section in its current form and do not need to invest work to restructure the content.

davidcarlisle:
[...]

PR #325, opened 2022-04-22
re-arrange chapter4 from @davidcarlisle, with operators appendix from @samdooley

davidcarlisle:
This rearranges the body of chapter 4 as suggested in issue #320 apart from the syntax tables which are collected in an appendix created by @samdooley as in PR #319  linking ids in both files updated so they link to each other, the body of the table is otherwise exactly as in PR #319.

It is still work in progress:

 *  the syntax tables need to be removed (or in some cases possibly changed to a simple reference to the syntactic class if that is not implied by the section 4.3.x heading)
 * <strike>There are several broken links (all flagged by respec so will be easy to find/fix</strike>
 * Many of the section 4.3.x sections now have multiple examples copied in from the original sections for each element. However we should review the examples and decide which ones to keep, or make new examples, to ensure all features covered in cases where it isn't simply repeating the same example, but with a different operator, in which case we can delete the duplicates.
 * Possibly should make examples folding as in chapter 3, but I suggest doing that at the end, it is easier to review the examples when not folded.
 * Slightly unrelated index.html has two css  changes that make the inline "two bar" mathml for content mathml not be shown, and similarly make the image version of the mathml rendering not be shown in chapter 4. (A permanent change here would change respec not to add them rather than using css to hide, but this allows easy testing)

Despite being incomplete as noted above, I think we could merge this at this stage (eg after discussion on the next call  on 2022-04-28) so the forks don't get too far ahead and lead to merge conflicts later.


NSoiffer (commented):
Looks fine to me. I'll let @sam or @brucemiller pull the trigger after one of them agrees.

davidcarlisle:
All respec issues have now been resolved so most of the items in the intitial description still to do but respec builds with no errors or warnings

brucemiller (approved):
[Note that I've focused on the @davidcarlisle's gh-pages, rather than the source]

I like where this is going.  Section 4.3 is still disturbingly long, but given the fact that the immense 4.4 is gone, that's not so bad. 

In the next iteration we may want to think about the order/organization of 4.3. It has 36 subsubsection! At the cost of having 5 levels, we might want Unary, Binary, Nary subsubsections?

Since there were recent commits, maybe more? I don't want to merge yet, but I think this is a good direction to go in.

davidcarlisle:
@brucemiller  yes I hope we can re-organise again and move everything up a level at some point (as well as dropping many of the examples)  The multiple subsections essentially map to the syntactic classes in the Relax NG, we could merge some of them if we wished, eg there is no structural difference between `binary-arith` , `binary-logical` and `binary-linalg` so they could all be merged to `binary` if wanted to lose all remnants of 4.4's "mathematical area" divisions.

brucemiller:
@davidcarlisle yes we could possibly merge some (sub)^n as well, but what I was thinking initially was to move (say) all the nary subsubsections as subsubsubsections under an nary subsubsection.

davidcarlisle:
> @davidcarlisle yes we could possibly merge some (sub)^n as well, but what I was thinking initially was to move (say) all the nary subsubsections as subsubsubsections under an nary subsubsection.

ah yes that might work, but after a merge:-) Probably should leave in this unmerged state until at least Thursday's zoom meeting,

src/character-set.html
```diff
@@ -242,7 +242,7 @@ <h4 id="chars_pseudo-scripts">Pseudo-scripts</h4>
    <p>In almost all mathematical contexts, pseudo-script characters should
    be associated
    with a base expression using explicit script markup in MathML.  For
-   example, the preferred encoding of <span>&#x201c;x prime&#x201d;</span> is</p>
+   example, the preferred encoding of <q>x prime</q> is</p>
```
davidcarlisle: @dginev  (or anyone else)  `character-set.html` only appears here as davidcarlisle/mathml/gh-pages was rebased against w3c so had the change from `<span>`  to `<q>` applied. The intention of the rebase was to make the files mostly the same and minimise diffs. And character-set.html in the gh-pages branches in w3c and davidcarlisle forks are in fact now identical.

It doesn't affect the actual merge, git will no doubt skip the "already applied commit", but I don't understand what the github UI diff is showing here?

davidcarlisle: This is resolved (and the diffs from that commit have gone)  (I used the fetch upstream GUI on my fork at gihub) That said I still don't understand what that diff is showing. commandline tools showed  (as far as I can tell) that the head of the gh-pages branch at w3c, davidcarlisle, and my local clones all had that `<q>` commit applied.

davidcarlisle:
> @davidcarlisle yes we could possibly merge some (sub)^n as well, but what I was thinking initially was to move (say) all the nary subsubsections as subsubsubsections under an nary subsubsection.

We discussed this on the call today, @brucemiller  like this?


I (re) added  a containing section for all n-aray (unary,binary) and moved the  nary-arith level sections under that.

This allows the duplicated paragraph of each of those sections to be removed and used as the intro paragraph to the new layer so for example:

https://davidcarlisle.github.io/mathml/#contm_nary

davidcarlisle:
https://github.com/w3c/mathml/pull/325/commits/cca9edd4aff0e6867740c542f20423c8faf6d229

@samdooley  @brucemiller  this commit does (just) the first three classes (n-ary arith,funcional,logical) but  I (hope:-) it is bringing Sam's and my arrangement closer together and giving a more unifiorm organisation across chapter 4 and appendices A, F and G.

Firstly this deletes the syntax table from 4.3.5.1 https://davidcarlisle.github.io/mathml/#contm_nary_arith  (as in Sam's PR)
But I add links to the combined table in appendix F and to the schema class in appendix A.

The links are in a standard W3C class="def(inition)"  paragraph which is a bit blue but that's just cosmetics

In order to have something to link to I adjusted the schema highlight javascript to (re) add IDs, and re-organised (this part of) the big table in appendix F, adding a grouping `<body>` for each of the `nary-xxx` and pulling the rows for each operator into the  relevant `tbody`,  Just three groups done but if completed would mean the order in chapter 4 and appendix F would match, also improved scope for compressing the appendix F table as all the entries in each group are essentially the same.

davidcarlisle:
I have rebased this PR branch over upstream/gh-pages so it is hopefully easier to see diffs against the w3c repo.

brucemiller:
As I've said, I like the direction this goes.  The diff is (necessarily) incomprehensible; I think the best way forward is to merge and proofread the material on its own merits (comparing to MML3, as necessary).

I like the grouping into sections (n-ary, binary, etc), although it does highlight a mismatch: we have what I might provocatively call "syntactic" n-ary (like `<plus/>` and "semantic" n-ary like `<sum/>` under the same heading. But that can be sorted out later, if it even needs to be.

davidcarlisle:
Merging as agreed on the call on 2022-05-05

PR #326, opened 2022-04-24
update example as suggested by @NSoiffer

davidcarlisle:
amend example of stretchy brackets, and regenerate TeX images (I decided to use png rather than gif)


NSoiffer (approved):
The images seem a little thin relative to the MathML renderings by Firefox, but they are much better at illustrating the point the example is trying to make than the earlier examples.

davidcarlisle:
@NSoiffer  yes I wasn't too happy with the images, but I'd spent too long fiddling with image magic's conversion settings, and was out of time, so decided they were good enough. I may try again using Stix as that is rather more robust than computer Modern and may survive better as an image.  (but that can be handled separately to his PR)

davidcarlisle:
@NSoiffer  remade images with Stix2, look a lot better

PR #331, opened 2022-04-26
Fixed up fraction examples (and changed the text so that they were al…

NSoiffer:
…l in the example section)

Added msqrt/mroot example section

Note: it seems like Firefox doesn't implement linethickness 👎

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -3890,17 +3890,7 @@
  
  <p>Thicker lines (e.g. <code class="attribute">linethickness</code>="thick") might be used with nested fractions;
  a value of "0" renders without the bar such as for  binomial coefficients.
- These cases are shown below:
- </p>
- <blockquote>
-  <p><img src="image/f3007.gif" alt="\binom{a}{b} \quad \genfrac{}{}{1pt}{}{\frac{a}{b}}{\frac{c}{d}}"></img></p>
- </blockquote>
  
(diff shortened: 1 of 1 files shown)


davidcarlisle (changes requested):
Currently neither rendering actually works, the bevelled one is marked as mml4p (ie, not core) the stacked fraction is currently incorrectly marked as core, but as you comment, it doesn't work in firefox.   We could mark it as `mml4p` but perhaps better would be to change `linethickness="thick"` to  `linethickness="1pt"` then it is valid as mathml core (and works in firefox)

Do you want me to generate latex (Stix Two) images for these ?

src/presentation-markup.html
```diff
@@ -3890,17 +3890,7 @@ <h5>Attributes</h5>
  
  <p>Thicker lines (e.g. <code class="attribute">linethickness</code>="thick") might be used with nested fractions;
  a value of "0" renders without the bar such as for  binomial coefficients.
```
davidcarlisle, line 3892: missing `</p>` here (respec doesn't mind, but I do:-)

davidcarlisle:
thanks for the update. If we don't add images we could probably discuss keeping at least the core example renderings in the exported spec.

davidcarlisle:
Merging this as linethickness fixed. If we are keeping the core rendering we will probably end up with fewer images generally so won't add images at this time.

PR #332, opened 2022-04-26
use q element

davidcarlisle:
The original MathML3 `xmlspec` XML sources had a `<quote>` element, but in the conversion to respec (which was a "creative merge" of the xmlspec source and its generated html)  this has got expanded out and is marked everywhere (over 100 instances) as 

`<span>&#x201c; ... &#x201d;</span>`

which isn't very maintainble source input. This PR is a simple global (almost) edit to change them all to the standard HTML

` <q> ... </q>`

(I left out contm-new for now as it would conflict with existing PR, if this PR lands I will make the same edit in the chapter 4 PR branch)

### DIFF
--- a/src/character-set.html
+++ b/src/character-set.html
@@ -242,7 +242,7 @@
    <p>In almost all mathematical contexts, pseudo-script characters should
    be associated
    with a base expression using explicit script markup in MathML.  For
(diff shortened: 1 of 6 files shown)


NSoiffer (approved):
These bothered me so I'm glad you made the change.

PR #333, opened 2022-05-02
Do not use removeOnSave so MathML in the exported final HTML

davidcarlisle:
Do not use removeOnSave so MathML is in the exported final HTML

Do not add inline MathML for Content MathML (mml4c) examples

Add altttext attribute (required for inline MathML in REC track documents)
generated from simple regex simplification of the example source.


davidcarlisle:
This currently inlines mml4p examples, we could either not do that, or we need to set up polyfills as part of this repo as we can't use external resources as far as I understand pubrules.

I (think) I would propose removing the mml4p examples on save, so just having them in the live respec draft, using links to the polyfill github, which can also be removed on save. That would just leave the mathml core examples in the exported REC HTML.

NSoiffer (commented):
As I understand it, this completely drops the output.

I understand the reasoning for not including external resources in the final spec, but without some output, the examples aren't very useful and there would be nowhere else for people to view them (besides looking at a different version of the spec).

Here's an idea: include static images in the spec. For the respec version, the polyfills are used and act both as a test of the polyfills and as an example of how they can be used. For the `spec.html` version, the static images are used.

Is this viable?

davidcarlisle:
@NSoiffer This PR (should?) make no change to images, any images that are there are still used. (separate suggestion to drop images in chapter 4, but looking mostly at chapter 3 here) The only real changes made by this PR are that

1. the inline mathml is kept in the exported version
2. content mathml is not inlined.

Your suggestion (having images and inline mathml in the respec version and just images in the exported) is basically the status quo, except that currently the poyfills are not loaded so we inline the mml4p and mml4c examples which don't work. We could add the polyfills and leave the exported version as-is with just images, but I am somewhat tempted to make the political statement of having inline mathml core rather than images in the normative spec for any examples that are in core.

Is there a simple script call to load all the polyfills? I could base something on 

https://mathml-refresh.github.io/mathml-polyfills/acid-test.html

but a simple

`<script src="https://mathml-refresh.github.io/mathml-polyfills/polyfills.js></script>`

would be nicer:-)

NSoiffer:
Sorry for not being clear. My suggestion is to include the images in the raw spec, but that respec doesn't show them, just the polyfill image. In the exported spec.html, it would use the images. I don't like the idea of showing both the live and the image. I think that would be confusing.

There isn't a script that loads them all, but of course, it would be simple to bundle up what is used:
```
    <link rel="stylesheet" type="text/css" href="[./menclose/menclose.css]>
    <script type="module">
        import { _MathTransforms } from './common/math-transforms.js'
        import './mglyph/mglyph.js'
        import './mfenced/mfenced.js'
        import './bevelled/bevelled.js'
        import './accent/accent.js'
        import './horiz-align/horiz-align.js'
        import './mathsize/mathsize.js'
        import './menclose/menclose.js'
        import './linethickness/linethickness.js'
        import './ms/ms.js'
        import './mpadded/mpadded.js'
        import './mtable/mtable.js'
        import './elem-math/elemMath.js'
        import './linebreaking/linebreaking.js'

        document.head.appendChild ( _MathTransforms.getCSSStyleSheet() );
        window.transform = function (elementId) {
            _MathTransforms.transform(elementId);
        }
```

FYI: the CSS for `menclose` to draw the arrowheads came from David Cervone and is _very_ clever.

davidcarlisle:
@NSoiffer  OK that makes sense, I'd like to merge this though (just handling the inline MathML) and I opened #334 issue to track experiments with handling the images.

davidcarlisle:
Merging as agreed on the call on 2022-05-05

PR #338, opened 2022-05-06
Compress n-ary relations section

davidcarlisle:
The repeated paragraphs for = < > <= >= are merged in to one combined paragraph

The paragraph describing neq deleted (that is already copied to its own section

Use of `"foo"` (which produces two right quotes) replaced by `<q>foo</q>`

The = example extended to be a chain of three, not just two

The image commented out (it is already set with css to not display, if we bring the images back, would need to be regenerated with new example)


brucemiller (approved):
Nice!

PR #339, opened 2022-05-06
compress arthmetic constants section.

davidcarlisle:
Combine descriptive paragraphs for each element into one.

Move all content and sample presentation examples under single heading for each type (so all can be folded togther)

Un-indented the small content examples so they become one line each

commented out the images (which have display:none anyway)


brucemiller (approved):
looks good

PR #350, opened 2022-05-09
Changes/fixes to "Script and Limit Schemata" section

NSoiffer:
Changes:
* Cleaned up text and removed (extra) images in the examples. Also removed the examples from text since they are now part of the display in the examples section.
* Fixed indentation of mmmultiscripts syntax example
* Added text about baseline alignment of the scripts

Note: the example with munderscripts to demonstrate differences is pretty subtle in Firefox and non-existant (currently) in Chrome. I'm not sure it is a great example. Perhaps we should use something different. Suggestions?

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -2974,9 +2974,9 @@
      </div>
  
     <p>The above should render as
-    <img src="image/f3001.png" alt="(\frac{a^2}{b^2})"class="vmiddle"></img>
+    <img src="image/f3001.png" alt="(\frac{a^2}{b^2})" class="vmiddle"></img>
     as opposed to the default rendering
-    <img src="image/f3002.png" alt="\left(\frac{a^2}{b^2}\right)"class="vmiddle"></img>.</p>
+    <img src="image/f3002.png" alt="\left(\frac{a^2}{b^2}\right)" class="vmiddle"></img>.</p>
  
     <p>Note that each parenthesis is sized independently; if only one of
     them had <code class="attribute">maxsize</code>=<code class="attributevalue">100%</code>, they would render with different
@@ -5800,9 +5800,9 @@
     </p>
  
     <p> Note that both scripts are positioned tight against the base as shown here
-    <img src="image/f3014.gif" alt="x_1^2"class="vmiddle"></img>
+    <img src="image/f3014.gif" alt="x_1^2" class="vmiddle"></img>
     versus the staggered positioning of nested scripts as shown here
-    <img src="image/f3013.gif" alt="x_1{}^2"class="vmiddle"></img>;
+    <img src="image/f3013.gif" alt="x_1{}^2" class="vmiddle"></img>;
     the latter can be achieved by nesting an <code class="element">msub</code> inside an <code class="element">msup</code>.</p>
    </section>
  
@@ -6116,7 +6116,7 @@
      </table>
  
  
-    <p>The difference between an accent versus limit is shown in the <a href="#presm_mover_ex">examples</a>:
+    <p>The difference between an accent versus limit is shown in the <a href="#presm_mover_ex">examples</a>.</p>
  
     <p>The default value of <em>accent</em> is false, unless
     <em>overscript</em> is an <code class="element">mo</code> element or an
@@ -6300,7 +6300,7 @@
     <p>This example shows the difference between nesting <code class="element">munder</code> inside
       <code class="element">mover</code> and using <code class="element">munderover</code> when
       <code class="attribute">movablelimits</code>=<code class="attributevalue">true</code>
-      and in <code class="attribute">displaystyle</code> (which renders the same as <code class="element">msubsup</code>).
+      and in <code class="attribute">displaystyle</code> (which renders the same as <code class="element">msubsup</code>).</p>
  
     <div class="example mathml mmlcore">
      <pre>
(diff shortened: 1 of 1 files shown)


src/presentation-markup.html
```diff
@@ -6118,14 +6116,7 @@ <h5>Attributes</h5>
...
  
-    <p>The difference between an accent versus limit is shown here:
-    <img src="image/f3018.gif" alt="\hat{x}"class="vmiddle"></img> versus
-    <img src="image/f3019.gif" alt="\hat{\strut x}"class="vmiddle"></img>.
-    These differences also apply to <q>mathematical accents</q> such as
-    bars <span>or braces</span> over expressions:
-    <img src="image/f3020.gif" alt="\overbrace{x+y+z}"class="vmiddle"></img> versus
-    <img src="image/f3021.gif" alt="\overbrace{\strut x+y+z}"class="vmiddle"></img>.
-    The MathML representation for each of these examples is shown below.</p>
+    <p>The difference between an accent versus limit is shown in the <a href="presm_mover_ex">examples</a>:
```
davidcarlisle: missing `#`:  `href="#presm_mover_ex"`

src/presentation-markup.html
```diff
@@ -6135,13 +6126,15 @@ <h5>Attributes</h5>
...
     <em>overscript</em> is an embellished operator, the <code class="attribute">accent</code> attribute of the <code class="element">mo</code>
     element at its core is used as the default value.</p>
+    <p>[[Mathml-Core]] does not support the <code class="attribute">accent</code> attribute on <a href="#presm_mo"></a>.
+      For compatibility with MathML-core, the <code class="attribute">accentunder</code> should be set on <code class="element">munder</code>.</p>
  
    </section>
  
    <section class="fold">
-    <h5>Examples</h5>
+    <h5 id="id="presm_mover-ex"">Examples</h5>
```
davidcarlisle: scrambled markup and `-` instead of `_`: `<h5 id="presm_mover_ex">`

src/presentation-markup.html
```diff
@@ -6135,13 +6126,15 @@ <h5>Attributes</h5>
...
  
-    <p>The MathML representation for the examples shown above is:</p>
+    <p>Two examples demonstrating how <code class="attribute">accentunder</code> affects rendering:</p>
```
davidcarlisle: they don't use `accentunder`, did you mean `accent` here ?

src/presentation-markup.html
```diff
@@ -6292,21 +6285,8 @@ <h5>Attributes</h5>
...
-    The MathML representation for this example is shown below.</p>
- 
-    <p>The difference in the vertical spacing is too small to be noticed on a
-    low resolution display at a normal font size, but is noticeable on a higher
-    resolution device such as a printer and when using large font sizes.  In
-    addition to the visual differences, attaching both the underscript and
-    overscript to the same base more accurately reflects the semantics of the
-    expression.</p>
- 
+    to the base and so that they follow the slant of the base as shown in the <a href="presm_munderover_ex">example</a>.</p>
```
davidcarlisle: missing `#` again

src/presentation-markup.html
```diff
@@ -6315,12 +6295,10 @@ <h5>Attributes</h5>
...
  
    <section class="fold">
-    <h5>Examples</h5>
+    <h5 id="presm_munderover_ex">Examples</h5>
  
-    <p>The MathML representation for the example shown above with the first
-    expression made using separate <code class="element">munder</code> and
-    <code class="element">mover</code> elements, and the second one using an
-    <code class="element">munderover</code> element, is:</p>
+    <p>This example shows the subtle difference between nesting <code class="element">munder</code> inside
```
davidcarlisle: I think it's too subtle to highlight (I'm not sure a difference is mandated at all in that case is it? I'd use a non subtle example, where it uses subscript layout

> This example shows the difference between nesting <code class="element">munder</code> inside
      <code class="element">mover</code> and using <code class="element">munderover</code>is particularly apparent when not in `displaystyle`, when the limits may be placed as for <code class="element">msubsup</code>`

```
<mstyle displaystyle="false">
 <mover>
 <munder>
  <mo>∑</mo>
  <mi>x</mi>
 </munder>
 <mi>n</mi>
 </mover>
 <mo>+</mo>
 <munderover>
  <mo>∑</mo>
  <mi>x</mi>
  <mi>n</mi>
 </munderover>
</mstyle>
```

davidcarlisle (changes requested):

src/presentation-markup.html
davidcarlisle: or probably with a more mathematically meaningful example, but the above does show the layout difference, firefox shows

image

src/presentation-markup.html
NSoiffer: Fixed

src/presentation-markup.html
NSoiffer: Fixed

src/presentation-markup.html
NSoiffer: Fixed

src/presentation-markup.html
NSoiffer: Good idea to show an example with moveable limits for this. I have changed the example to use `i` instead of `x` since `n` is usually an integer.

src/presentation-markup.html
NSoiffer: Fixed

davidcarlisle:
I pushed a couple of markup fixes but looking good now thanks, merging.

PR #351, opened 2022-05-10
spelling corrections and removal of leftover(?) broken-ish spans

NSoiffer:


NSoiffer:
Forgot to switch branches for the fix to 335 -- shouldn't cause any problems though.

davidcarlisle (approved):
yes the spans were the result of mathml3 change markup. The safest thing to do originally was to delete the `diff="chg"` markup and leave the `<span>` as that saved finding the matching `</span>` but I've since deleted a lot of them as they come up.

PR #353, opened 2022-05-10
merged nary-arith

davidcarlisle:
This section actually expanded slightly to separately describe plus/times and gcd/lcm


brucemiller:
looks fine.

PR #355, opened 2022-05-11
Tabular math fixes

NSoiffer:
Remove allowed unitless value mention and then add mention that it was allowed in MathML 3 but deprecated in 4 (two places)

A few wording tweaks

Remove some images/<pre>s from the examples.

I only did a cursory review of the Alignment Markers section since that may be deprecated.

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -2303,11 +2303,13 @@
         <td colspan="3" class="attdesc">
    Specifies an additional indentation offset relative to the position determined
    by <code class="attribute">indentalign</code>.
-   <span>When the value is a percentage value or number without unit,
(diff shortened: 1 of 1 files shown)


src/presentation-markup.html
```diff
@@ -7949,16 +7933,7 @@ <h5>Inheritance of <code class="attribute">groupalign</code> values</h5>
  <h5>MathML representation of an alignment example</h5>
  
  <p>The above rules are sufficient to explain the MathML representation
- of the example given near the start of this section.
```
davidcarlisle, line 7952: </p> here but I pushed a commit adding this

davidcarlisle (approved):

Issue #354, opened 2022-05-11
Fix examples at the start of "3.5.5 Alignment Markers"


NSoiffer:
The examples at the start of 3.5.5 are done with a `pre` and don't use italics for the vars or proper math spacing. These should be images (there are not any polyfills for the alignment markers).

The same needs to be done with the examples in https://w3c.github.io/mathml/spec.html#table-cells-that-are-not-divided-into-alignment-groups.

davidcarlisle:
or perhaps clearer than images here espcially as it's introductory "somewhat like that shown here:" rather than claiming to show MathML, would be to use HTML tables.

davidcarlisle:
these now use html table markup, closing

PR #357, opened 2022-05-12
Tabular

NSoiffer:
Change the examples in the alto use fake MathML with tables as suggested by @davidcarlisle in https://github.com/w3c/mathml/issues/354

The first example in #table-cells-that-are-not-divided-into-alignment-groups is not right though. The long first cell is spacing out the rest of the cells. I played with `table-layout` but didn't see a solution there. Maybe something with width and overflow might work. @davidcarlisle: do you know how to get this right?

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -2303,11 +2303,13 @@
         <td colspan="3" class="attdesc">
    Specifies an additional indentation offset relative to the position determined
    by <code class="attribute">indentalign</code>.
-   <span>When the value is a percentage value or number without unit,
+   When the value is a percentage value,
    the value is relative to the
    horizontal space that a MathML renderer has available, this is the current target
    width as used for
-   linebreaking as specified in <a href="#presm_linebreaking"></a></span>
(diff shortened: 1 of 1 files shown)


davidcarlisle (changes requested):
I pushed some general HTML validity commits, so approve modulo the (now) duplicated table example.

I added a second version with the long heading overflowing left and right

image


But I'm not sure that's the layout you intended, in either case, I left in the original table so one of them needs **removing** before this is merged.

The third table shown here with the short heading has a left aligned heading, which I'm not sure fits with the preceding description "`using the default columnalign="center"`" should that be centred? If it should be left aligned, presumably the long heading example should just overflow on the right, not left and right as I added here?

NSoiffer:
Good catch. I was just trying replicate what was there, but what was there was wrong. Looking at the V3 spec, indeed the math should be centered. In the second example, the header should be centered.

I'm not thrilled with your solution of padding with `&nbsp;`, but I don't have a better idea. I'll adjust the examples.

NSoiffer:
FYI: the short heading on the last example doesn't look centered, but it really is. Sort of an optical illusion.

PR #348, opened 2022-05-07
use text not images in lambda share paragraph

davidcarlisle:
The inline images for math just look horrible:-)

I left this in the default sans serif (as done for small inline math terms elsewhere) we could set up a serif style for inline math-as-text but I think it's OK like this? 

old

image

new

image

PR #360, opened 2022-05-15
folding 4.3.x nary-stats

davidcarlisle:
The last of the subsections of 4.3

If this along with #348,  #358  and #359 are merged, all "Sample Presentation" examples sections in chapter4 are folded.

### DIFF
--- a/src/contm-new.html
+++ b/src/contm-new.html
@@ -4428,13 +4428,64 @@
      <p>The <code class="element">mean</code> element represents the function returning arithmetic mean or average of a
      data set or random variable.</p>
 
-     <div id="s_data1.mean.ex1" class="mathml-example">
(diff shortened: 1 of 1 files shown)


brucemiller (approved):
Overall looks good; hard to judge the code samples w/o previewing, but hopefully we'll catch any errors when merged.

davidcarlisle:
> Overall looks good; hard to judge the code samples w/o previewing, but hopefully we'll catch any errors when merged.

@brucemiller  thanks yes you need to run a local server really to see. The line based diff with these are particularly misleading as moved mathml source examples all look the same to diff so it shows diffs showing content mathml changing line by line to presentation, when in fact unchanged examples were really just re-ordered to allow better folding. But if all 4 of these can be merged in hopfully the spec will be looking better before tomorrow's call.

PR #363, opened 2022-05-19
Davidc/intent

davidcarlisle:
Initial steps...

Adds new subsections to 3.1.9 adding MathML Core and intent attributes (following status quo of intent being Presentation only)

Start cleaning up chapter5, just show one version of the grammar, can now link to the earlier section introducing intent  so deleted the ednote querying whether that would be in chapter 3 for just presentation or chapter 2  for global.

Still needs more examples and more words.


NSoiffer:
Seem to look good (a little hard to tell without seeing it). I'll merge and double check.

PR #364, opened 2022-05-19
comment out mime appendix inclusion,  add new bibref

davidcarlisle:
commented out the inclusion of the appendix and added a new bibref item for the separate document.  Changed one internal  link  in  chapter 6 to  a bibref.


NSoiffer (approved):
This seems good, although more might need to be done on Chapter 6.

PR #366, opened 2022-05-22
Davidc/intent2

davidcarlisle:
Starting to add a bit more structure to chapter 5, still mostly just adding editorial notes and stub text but replacing some of the current text that makes no gramatical sense as it was just copied without context from earlier documents.

I suggest that this is merged more or less as is, even if none of his text will survive to the public draft, to allow more people to see/comment on intent descriptions.

@NSoiffer I added some mathcat generated text but I may not have chosen the best options.... (I seem to have used clear speak / verbose / plain)

### DIFF
--- a/index.html
+++ b/index.html
@@ -87,7 +87,7 @@
 	    m.setAttribute("class",
 			   mclass.replace("example mathml","") +
 			   " inlinemath");
-	    m.innerHTML=mmllist[i].textContent;
+	    m.innerHTML=mmllist[i].querySelector("pre").textContent;
 	    var atext;
 	    var aattrib=mmllist[i].getAttribute("alttext");
 	    if (aattrib)  {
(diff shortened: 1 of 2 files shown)


NSoiffer:
I'll merge this, but I'm not a fan of using nonsense examples. I think breaking the examples up and dropping the right arrow would be better. Alternatively, you could put them in a list. At least that isn't mathematical nonsense. Although it uses more space, I think breaking them up or using a table would be better.

davidcarlisle:
@NSoiffer  yes I agree really, especially after I added the readings which was a later idea. Originally it just seemed a convenient way to show  the comparison more compactly and it is hard to have a real mathematical example showing the same syntax used for muliple conflicting meanings inline. A list or table may be better, as you say.

NSoiffer:
One thought I had last night was to combine two into one. E.g, f'(x^n) or
D^{(n)} sin(x^n). Transpose would be separate but maybe you could do
something like A^H = A^* = A^T (hermitian, complex-conjugate, transpose)
although I'm not sure whether people would just say "star" or "complex
conjugate" for the middle one, so may leave that out.

PR #367, opened 2022-05-22
first attempt at referencing polyfills

davidcarlisle:
@NSoiffer  This adds the combined polyfill to the respec post filter and it seems to mosly work, but I guess it has lost some CSS along the way, the alignment in `<mscarries` is off and `<msline` isn't appearing,.

The console shows

GET `http://localhost:8080/w3c/mathml/elem-math/elemMath.css` is 404

but that CSS comes from inside the polyfill JavaScript so no sure how best to get it to use 
https://mathml-refresh.github.io/mathml-polyfills/elem-math/elemMath.css
I tried adding a link to that "from the ouside" but without any effect.

I think I got as far as I can go, can you pick his up, either pushing to this branch, or merging anyway and then fixing in a later PR?

the mscarries example looks like this:

image

### DIFF
--- a/index.html
+++ b/index.html
@@ -86,7 +86,6 @@
         sorttable.makeSortable(document.getElementById('oper-dict-table'));
     }
     function mmlpolyfill () {
-	document.querySelectorAll("div.mml4p").forEach((div) => transform(div));
 	document.querySelectorAll("math.mml4p").forEach((div) => transform(div));
     }
     function showcore (){
--- a/index.html
+++ b/index.html
@@ -7,6 +7,14 @@
       src='https://www.w3.org/Tools/respec/respec-w3c'
       class='remove'>
   </script>
+  <script type="module" class='remove'>
+        import { _MathTransforms } from 'https://mathml-refresh.github.io/mathml-polyfills/all-polyfills.js'
+
+        document.head.appendChild ( _MathTransforms.getCSSStyleSheet() );
+        window.transform = function (elementId) {
+            _MathTransforms.transform(elementId);
+        }
+  </script>
   <script src="sorttable.js"></script>
   <script class='remove'>
     function relaxhl (r,s,u) {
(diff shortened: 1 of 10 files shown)


NSoiffer:
Do an update. I think I fixed that a few days ago. I just did another fix or two today.

davidcarlisle:
> Do an update. I think I fixed that a few days ago. I just did another fix or two today.

I was referencing https://mathml-refresh.github.io/mathml-polyfills not my local copy so should have had the current version, but I just saw the recent commits in polyfill, so I'll test this branch again....

davidcarlisle:
@NSoiffer I have updated everything here and at polyfills, and rebased this branch against main, but the CSS doesn't load for me. Perhaps you could merge, then use the main gh-pages rendering for testing?

davidcarlisle:
@NSoiffer  It works locally if I force loading the CSS by absolute URL with

```
diff --git a/elem-math/elemMath.js b/elem-math/elemMath.js
index 91b408d..ef74a06 100644
--- a/elem-math/elemMath.js
+++ b/elem-math/elemMath.js
@@ -977,7 +977,7 @@ function addStyleSheetToShadowRoot(shadowRoot) {
     const link = document.createElement("link");
     link.rel = 'stylesheet';
     link.type = 'text/css';
-    link.href = './elem-math/elemMath.css';
+    link.href = import.meta.url + '/../elemMath.css';
     style.appendChild(link);
     shadowRoot.appendChild(style); 
 }
```

davidcarlisle:
Note github PR show all the rebased commits, this PR only actually changes `index.html` see the "double dot" diff

https://github.com/w3c/mathml/compare/main..davidc/polyfill?expand=1

More details on `..` v `...` at

https://blog.developer.atlassian.com/a-better-pull-request/

NSoiffer:
I'll do the merge and then we can try to figure out what is going on. There are people in the group more versed in CSS than I am. I'm sure there's a simple way to fix it.

PR #369, opened 2022-05-24
remove non core rendering from exported HTML

davidcarlisle:
adds `removeOnSave` class to math generated from `mml4p` examples so they are removed from exported HTML

re-instate some images, add two new ones for e=mc^2 mlabeledtr (which is a firefox rendering) and maligngroup (latex stix2 rendering)  

As the images are explicit ``  from around each image that came from the initial conversion to respec


NSoiffer:
Looks ok

PR #371, opened 2022-05-26
move repeating decimal images after mathml 

davidcarlisle:
Rearranging examples 3.6.8.4 to follow the usual layout with the four images split up and placed one under each example rather than all together at the start.

This means they come next to the inserted polyfill renderings in the respec draft and means we can drop the awkward phrase

> The MathML for the preceding examples above is given below.


NSoiffer (approved):
I'm doing some rewrites in that section, but for now, this is fine.

PR #372, opened 2022-05-26
core and open intent lists

davidcarlisle:
Probably not final wording but trying to keep the current text in line with today's call.

Document two lists, **core** and **open**

Start to have some words about normalising names before looking up in the intent lists


dginev (approved):
Very nice!

PR #373, opened 2022-05-27
Review and rewrite section 4.1 introduction

samdooley:
Complete rewrite for section 4.1 introduction to track changes in other parts of the document.

GitHub pages view should be visible at:
https://samdooley.github.io/mathml/#contm_intro

### DIFF
--- a/src/contm-new.html
+++ b/src/contm-new.html
@@ -1,168 +1,212 @@
 <section>
  <h2 id="contm">Content Markup</h2>
-<!-- <div class="issue" data-number="273"></div> -->
  <div class="issue" data-number="284"></div>
  <section>
   <h3 id="contm_intro">Introduction</h3>
 
   <section>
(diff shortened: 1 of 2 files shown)


davidcarlisle (approved):
This looks good to me, I have read the diff, and then side by side the two github pages views. Marking as approved but leaving it unmerged to give others chance to look/comment, but you can merge whenever you are ready,

NSoiffer (approved):
> The defined set of content elements is designed to be adequate for simple coding of formulas typically used from kindergarten through the first two years of college in the United States, that is, up to A-Level or Baccalaureate level in Europe.

That has always been, and remains, dubious. Is there a way to rewrite this so it is more truthful (without having to say "we missed a bunch" in previous versions)?

Other than that, it looks good. I'll do the merge and we can circle back and argue about that statement for a few week ;-)

Issue #362, opened 2022-05-19
Inline math in the spec


davidcarlisle:
At various points in the document we have small inline expressions. In MathML3 we used the document italic (`<i class="var">`)  unless there were any fractions or Greek in which case we used vertically centred images.

<div class="notinspec">

Alternatives seem to be:

* Use inline images
* Use document text sans serif italic (sloped)  
* Use document serif Italic ie css style `.var` to be (any?) serif font
* Use Unicode Math Italic Block  (possibly using the reader's default font for this range)
* Use MathML Core

The current draft is using the second of these ( `<i` giving sans serif italic). I think that's fine for casual expression references in a sans serif document, but opening the issue so we can decide and treat them all the same way.

```
src/contm-new.html:<i class="var">H<
src/contm-new.html:<i class="var">e<
src/contm-new.html:<i class="var">H<
src/contm-new.html:<i class="var">e<
src/contm-new.html:<i class="var">H<
src/contm-new.html:<i class="var">e<
src/contm-new.html:<i class="var">H<
src/contm-new.html:<i class="var">e<
src/contm-new.html:<i class="var">H<
src/contm-new.html:<i class="var">e<
src/contm-new.html:<i class="var">H<
src/contm-new.html:<i class="var">e<
src/contm-new.html:<i class="var">H<
src/contm-new.html:<i class="var">e<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">y<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">y<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">y<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">f<
src/contm-new.html:<i class="var">f<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">y<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">y<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">vector(complex)<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">y<
src/contm-new.html:<i class="var">cdbase<
src/contm-new.html:<i class="var">cd-name<
src/contm-new.html:<i class="var">symbol-name<
src/contm-new.html:<i class="var">x<
src/contm-new.html:<i class="var">y<
src/contm-new.html:<i class="var">x<
[...]

PR #374, opened 2022-05-27
inline MathML in the document source

davidcarlisle:
This is a _draft_ PR with gh-pages view visible at

respec version:

https://davidcarlisle.github.io/mathml/index.html

exported HTML:

https://davidcarlisle.github.io/mathml/spec.html

At the call on 2022-05-26 there was some discussion of the "html-math" fragments in the spec.

This PR adds another JavaScript filter which runs on save (so makes no difference in the respec version) the exported HTML gets a new `<script>` element.  This script:

* Tests if MathML Core rendering is available (height of  `<msqrt><mi></mi></msqrt>` > 0)
* If It is available nothing happens
* If Core is _not_ available
    * all `<math class="mmlcore">` which have been generated from Core examples are removed
    * all  `<math alttext=".." class="doc">` which have been placed as MathML markup in the source are replaced by their `alttext` attribute (slightly modified with a regex eg to make single letter variables italic)

So the intention, if we decide to do this, would be to modify `index.html` as here and then use 

```
<math alttext="..." class="doc">...</math>
```

wherever needed in the source.

I did one section as an example here, compare the lambda-expression paragraphs at

https://davidcarlisle.github.io/mathml/spec.html#contm_share_binding

In a MathML-Core browser and a browser without. Also compare the current version using html directly

https://w3c.github.io/mathml/spec.html#contm_share_binding

I was planning not to touch chapter 4 as it is with  @samdooley but this is a draft PR not planning to merge as-is, and that section is one of the few sections that makes real use of inline math in paragraph text.

Almost all other cases of inline math (search for `<i class="var">`) can I think  be removed rather than being replaced by MathML as they are essentially just describing an existing MathML example so for example in chapter  3 I made again one sample change, at

https://davidcarlisle.github.io/mathml/spec.html#presm_invisibleops

The fourth column which had "example use" of invisible operators I deleted as the MathML versions would be identical to the inserted MathML that is generated from the immediately following examples.

Whether or not we add inline MathML, it may be worth using the filter here as removing the rendering of the MathML examples looks better than having the "green boxes" just highlighting broken rendering. 

Sorry about the length of this PR comment:-)


NSoiffer:
FYI: this is the speech rendering for the one with MathML (which I think looks better also) using MathCAT:
This represents a term 
> lamda x dot f of open paren lamda x dot g of x comma g of x close paren 
> which has two sub-terms of the form 
> g of x
> one with 
 
 This is the speech with the HTML Math (rendering would vary with different punctuation settings):
> This represents a term λ x dot f λ x dot g x g x which has two sub-terms of the form g x one with

NSoiffer:
Look good, although it seems like there should be more inline math.

davidcarlisle:
@NSoiffer Yes as noted above, ths PR lands the main supporting javascript and css, plus a couple of sample sections.

The full list of inline math is in the linked issue #362 they can be picked off and deleted, left or changed to

`<math alttext="..."class="doc">...` 

as part of normal edit cycles, now this PR has landed the support code.

Issue #376, opened 2022-05-30
intent grammar -- should allow '.12' as a number


NSoiffer:
Currently, the spec says
```
number   := '-'? digit+ ('.' digit+)? 
```

But this does not allow a leading decimal point. For example:  `.1`. That seems like a mistake.

I suggest:
```
number  := '-'? (digit+ '.'? digit* | '.' digit+)
```
Or to just modify what we already have:
```
number  := '-'? (digit+ ('.' digit+) | '.' digit+)
```
But some regex handlers (such as Rust's) don't support nested parens (it's a speed reason), so I think the first form is a better one to use in the spec.

dginev:
This is already possible by using `0.1` and `0.12`.

If we appear to have enough energy for extending the grammar for numeric literals, I think I will have to re-open my plea for motivating examples for each class of additional syntactic coverage and its benefits to AT. Starting with the negative numbers, since we skipped over them rather quickly.

When would we need `.1` marked up explicitly as `intent=".1"` ? Couldn't we just assume it will already be instantiated via an `<mn>.1</mn>` that would need no further annotation? Having explanations for this would also help intent adopters down the line.

NSoiffer:
Off the top of my head, I don't have an example that involves numbers in any way. But if we are going to allow numbers, we shouldn't restrict them. Going to an extreme using your argument: I can't think of a reason to allow all of NCNAME. In particular, I can't give you an example that starts with "zz" or a Roman letter followed by a Greek one, so why are we allowing them? I realize you said "syntactic class", but is ".12" and "0.12" really different syntactic classes? It is just capturing what is typically used with we talk about numbers with decimal points.

dginev:
> I can't think of a reason to allow all of NCNAME. [...]  so why are we allowing them?

I was told in our meeting last Thursday that NCName was the only "politically viable" design choice that does not enter a very painful discussion of what exactly a "letter" is/ought to be for `intent`. My initial personal preference was for having a more restricted `name` than NCName, which only relies on a Unicode-friendly category of "letters" and the "-" separator. It still is, if it ever becomes "politically viable".

> Off the top of my head, I don't have an example that involves numbers in any way.

Right - that is my concern - neither do I. Early on it seemed sensible to allow some basic numeric literals for "exceptional cases" where one may want to write them inside an intent expression. Though I can't come up with a great example at the moment.

[...]

PR #377, opened 2022-05-31
use exising number definiton from chapter2

davidcarlisle:
This removes the ad-hoc definition of `number` from the intent grammar and instead references the existing definition in chapter 2

https://w3c.github.io/mathml/#syntax-notation-used-in-the-mathml-specification

That definition is based on unsigned-number and defers to CSS for the fine print.

This also addresss issue #376

image

### DIFF
--- a/src/fundamentals.html
+++ b/src/fundamentals.html
@@ -189,7 +189,7 @@
       </tr>
 
       <tr>
-       <td id="type_number"><em>number</em></td>
+       <td id="type_number"><dfn><em>number</em></dfn></td>
        <td>
         an optional prefix of "-" (U+002D), followed by an <a href="#type_unsigned-number">unsigned number</a>,
        representing a terminating decimal number (a type of rational number)</td>
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -108,12 +108,11 @@ the user and agent.</p>
 <pre class="def bnf">
 intent   := number | NCName | argref | function
 function := (NCName | argref) '(' intent [ ',' intent ]* ')'
-number   := '-'? digit+ ('.' digit+)? 
 argref   := '$' NCName
 </pre>
 
 <p>Here <a href="https://www.w3.org/TR/REC-xml-names/#NT-NCName"><code>NCName</code></a>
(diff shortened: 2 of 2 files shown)


dginev (commented):
Why not use the exact [CSS number](https://www.w3.org/TR/css-values-3/#numbers) as the intent `number`, without passing through the extra layer of indirection? An example difference would be that the positive [signed zero](https://en.wikipedia.org/wiki/Signed_zero) `+0` would be valid by the CSS definition, but not by the current MathML definition.

Probably worth at least a mention: It appears the CSS definition also brings in "E notation", so e.g. `-1e-1` will also be a valid number for intent with this PR.

NSoiffer (commented):
[Update: @dginev commented while I was writing my comment, so somewhat repetitive]

CSS defines it as:
> When written literally, a number is either an [integer](https://www.w3.org/TR/css-values-3/#integer), or zero or more decimal digits followed by a dot (.) followed by one or more decimal digits and optionally an exponent composed of "e" or "E" and an integer. It corresponds to the [<number-token>](https://www.w3.org/TR/css-syntax-3/#typedef-number-token) production in the [CSS Syntax Module](https://www.w3.org/TR/css-syntax/) [[CSS3SYN]](https://www.w3.org/TR/css-values-3/#biblio-css3syn). As with integers, the first character of a number may be immediately preceded by - or [+](https://www.w3.org/TR/selectors4/#selectordef-adjacent) to indicate the number’s sign.

So this add exponent notation.  I'm not a big fan of that as it now adds complication as to how to pronounce the number, although I suppose there are many ways to pronounce numbers. E.g, 1022 could be "one thousand twenty two" or "one thousand and twenty two" or even "ten twenty two".

It also adds `+` as a possible starting char -- that doesn't worry me though. Just pointing out the change.

davidcarlisle:
@dginev  that's a good question. I think  when edited chapter 2 I was trying to stay close to the mathml3 chapter2 table but add css or core references where it made sense. 

Here I think this PR is right to make the intent grammar reference chapter2 but I'm not sure the chapter 2 definition is right: css allows 1e2 notation (and leading +)  which we did discuss earlier and rejected at the time.

Out of time today. sorry.

PR #382, opened 2022-06-06
elem math changes -- moved the images into "mml4p" tags, but as comments

NSoiffer:
Moved images to inside the `<div class="example mathml mml4p">`.

They are currently commented out but potentially they can be uncommented and used in the non respec version.

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -8385,11 +8385,11 @@
       &lt;msline/&gt;
     &lt;/mstack&gt;
  </pre>
- <!-- <blockquote>
-  <img src="image/em-plus.png" alt="\begin{array}{r}   424 \\   +33 \\   \hline \end{array}"/>
-  </blockquote> -->
  </div>
- 
+ <blockquote>
+  <img src="image/em-plus.png" alt="\begin{array}{r}   424 \\   +33 \\   \hline \end{array}"/>
+ </blockquote>
+
  <p> Many more examples are given in <a href="#presm_elemmath_examples"></a>.</p>
  
  <section>
@@ -9119,11 +9119,11 @@
       &lt;msline/&gt;
     &lt;/mstack&gt;
  </pre>
- <!-- <blockquote>
-  <img src="image/em-plus-shift.png" alt="\begin{array}{@{}r@{}}    424 \\   +\phantom0 33 \\   \hline \end{array}"/>
-  </blockquote> -->
  </div>
- 
+ <blockquote>
+  <img src="image/em-plus-shift.png" alt="\begin{array}{@{}r@{}}    424 \\   +\phantom0 33 \\   \hline \end{array}"/>
+  </blockquote>
+
  <p>The next example illustrates how to put an operator on the right. Placing the
  operator on the right is standard in the Netherlands and some other countries.
  Notice that although there are a total of four columns in the example,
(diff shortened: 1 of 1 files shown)


davidcarlisle:
hmm the styling currently assumes the img blockquote will be immediately after the pmml4 div not inside eg

https://w3c.github.io/mathml/#example-1

image


If I uncomment one of the new commented imge to try it makes

image

Apart from that I agree having them in a consisent place not in a random earlier paragraph is good.

If we don't want the image in the respec version, I think the simplest is to have the blockquote after  the div with class, say, `pmml4` then it just takes a css rule setting `blockquote.pmm4` to be `display:none` and remove the css rule on export which is easier than uncommenting comments. We could do that for all the examples, not just elem-math?

I think

> Below is the example shown at the start of the section: the digits inside...

Needs to be

> In the first example, the digits inside....

or some such as the image is nolonger shown at the start of that section.

NSoiffer:
I thought it would be easier to style since it would be simply `mml4p .blockquote` No need to give the blockquotes a class. Also, structurally, that makes it clear that the image is tied to the MathML.

However, if that's not the case, I'm good with changing the location. Let me know I will adjust all the examples in the presentation section.

I know you feel some renderings from the polyfills are poor. Let me know which you feel are particularly bad and I'll work on them. If I can't make them better, then we can go with images. A question will be whether they need to render well in all browsers. In particular, ones that don't follow MathML core, especially wrt respecting CSS on MathML.

Also, the referenced example is at the start of the section but I apparently missed moving the image inside the mml4p. I'll adjust it as per whatever you feel is best for the rest of the images.

davidcarlisle:
I meant it's easier to css style them than auomatically uncomment them,  not that it is easier to be nested/not-nested.

However the nested version (without more css)  produces a double left bar and double two-colour background panel, which looks odd to me.

Actually I'd leave things as in

https://w3c.github.io/mathml/#example-1

wih image and polyfill in respec version and just the image in spec.html 

If you would rather the images are hidden in respec we can do that of course as well, although there  is a danger that things that are hidden from the editors draft don't get proof read until it is too late, espcially if any examples are edited.

NSoiffer:
I changed the structure, but I don't like the look of repeating the display twice. Even though I know why, it's confusing to see it twice. The color coding may be meaningful or memorable to some people, but I imagine most causal readers/readers trying to check one thing would be confused.

Don't you want a class on the image (e.g, `mml4p`) you can distinguish which ones are repetitive if the group decides they should be suppressed in the respec version?

As for missing something -- I think we just need to be diligent to check the `...spec.html` version if one of the images is changed.

If I forget, please bring this up at the meeting and see if anyone has a suggestion for what to do.

davidcarlisle:
> I changed the structure, but I don't like the look of repeating the display twice. Even though I know why, it's confusing to see it twice. The color coding may be meaningful or memorable to some people, but I imagine most causal readers/readers trying to check one thing would be confused.
> 
> Don't you want a class on the image (e.g, `mml4p`) you can distinguish which ones are repetitive if the group decides they should be suppressed in the respec version?
> 
> As for missing something -- I think we just need to be diligent to check the `...spec.html` version if one of the images is changed.
> 
> If I forget, please bring this up at the meeting and see if anyone has a suggestion for what to do.

@NSoiffer  OK I addd one line of CSS so the images are hidden in respec (and the CSS is removed on save, so they appear in the exported spec)

https://w3c.github.io/mathml/#presm_repeatdec

just has the polyfill and

https://w3c.github.io/mathml/spec.html#presm_repeatdec

just has the image.

The ones hidden like this are `<blockquote class="mml4p">`  I only added this class if the image was not referenced in the text and appears immediately after the example `<div>` (so it is at the same point as the inserted polyfill rendering)

So repeated decimal as above but the addition example

http://localhost:8080/w3c/mathml/index.html#presm_addsub

has an image followed by

> The MathML for this is:

followed by the mathml source followed by the polyfill rendering, so you can't simply swictch between the poyfill and the image.

I would be easy I think to make them all be in the new style, with some minor rewording and re-ordering.

davidcarlisle:
ah sorry I meant to merg this _first_ I'll clean up...

davidcarlisle (approved):

Issue #346, opened 2022-05-07
Appendix D (Accessibility): needs text


davidcarlisle:

Steve-Noble:
If we wanted to follow the example of the [EPUB 3 Spec](https://www.w3.org/TR/epub-33/), we could add a brief section called "Accessibility" (in the EPUB spec, it is Section 10), and then point to a separate MathML Accessibility specification which lives independent from the full MathML spec. In that case we would not need to have Accessibility as an appendix.

I provide some text for such a brief section below. However, if there is opposition to this approach, then the text below can be repurposed for the intro of the accessibility appendix, if we must really go to that approach. Nonetheless, having "MathML Accessibility" as its own independent specification may have several benefits.

Here's the text...

### Accessibility

As an essential element of the Open Web Platform, the W3C MathML specification has the unprecedented potential to enable content authors and developers to incorporate mathematical expressions on the web in such a way that the underlying structural and semantic information can be exposed to other technologies. Enabling this information exposure is foundational for accessibility, as well as providing a path for making digital mathematics content machine readable, searchable and reusable.

The internationally accepted standards and underpinning principles for creating accessible digital content on the web can be found in the W3C's Web Content Accessibility Guidelines (WCAG) [[wcag2](https://www.w3.org/TR/WCAG2/)]. In extending these principles to digital content containing mathematical information, WCAG provides a useful framework for defining accessibility wherever MathML is used.

As the current WCAG guidelines provide no direct guidance on how to ensure mathematical content encoded as MathML will be accessible to users with disabilities, a separate specification, MathML Accessibility [mathml-a11y], defines how to apply these guidelines to digital content containing MathML.

This specification recommends that digital content containing MathML conform to the accessibility requirements defined in [mathml-a11y]. A benefit of following this recommendation is that it helps to ensure that digital mathematics content meets the accessibility requirements already widely used around the world for web content. In addition, ensuring that digital mathematics materials are accessible will expand the readership of such content to both readers with and without disabilities.

#### NOTE
This specification does not integrate the accessibility requirements to allow them to adapt and evolve independent of the MathML specification — accessibility practices often need more frequent updating. The accessibility specification is also intended for use with past, present, and future versions of MathML, in addition to considerations for both the MathML-Core and the full MathML specification. The approach of a separate specification ensures that the evolution of MathML does not lock accessibility in time, and allows content authors to apply the most recent accessibility requirements.

NSoiffer:
[...]

PR #386, opened 2022-06-13
suggested possible rewording of text from #346

davidcarlisle:
suggested possible rewording of text from #346 mostly using respec link markup, and rewording not to imply a separate document

suggested as a PR to get some initial text into the document, but feel free to ignore if you would rather edit directly

### DIFF
--- a/src/accessibility.html
+++ b/src/accessibility.html
@@ -1,6 +1,48 @@
-<section class="appendix informative">
+<section class="appendix">
  <h2 id="accessibility">MathML Accessibility</h2>
     <div class="issue" data-number="346"></div>
     <div class="issue" data-number="282"></div>
 
+ 
+<p>As an essential element of the Open Web Platform, the W3C MathML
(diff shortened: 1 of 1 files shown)


NSoiffer (commented):
I don't think the last paragraph is very good. The notes can't have requirements. It could be "best practices". That fits with "accessibility practices often need more frequent updating".  It needs other tweaks also.

davidcarlisle:
> I don't think the last paragraph is very good. The notes can't have requirements. It could be "best practices". That fits with "accessibility practices often need more frequent updating". It needs other tweaks also.

Yes. To be honest I wouldn't expect any of the text to survive more than a couple of rounds of edits. The main purpose of this PR is to kick start the process on this appendix. Currently we are over half way through the charter and not a single word has been added.

NSoiffer:
Ok. Let's get something in and hopefully @Steve-Noble will have something to replace it with in the next few days.

PR #389, opened 2022-06-20
[Intent grammar] readability touchup, restricted numbers

dginev:
We had productive consensus in the meeting on June 9th about extending the intent grammar with empty arguments and higher-order applications, as seen in  [David's commit](https://github.com/w3c/mathml/commit/a2a141de31cccb37582df98b6f02d7a174685df7) shortly after.

I was traveling, so couldn't contribute much as to the specific grammar aesthetics. Trying that with this PR here, motivated by:
 * trying to avoid abbreviations that may be hard to interpret by novices
   * finding meaningful names that aren't misleading is hard in general. Still, I took some inspiration from [lambda calculus](https://en.wikipedia.org/wiki/Lambda_calculus) naming, and replaced the "funcall" abbreviation with the single word "application".
   * similarly, I moved the emphasis for `argref` to the full word `reference` since we have a single reference, but confusingly two possible understandings of the word `argument` -- one for the `arg` attribute, another for the argument slot in an "intent expression".
   * last, I added a level of indirection to `NCName` and explicitly called it `name`, to convey it's really meant for naming concepts. I was torn between `name` and `concept`. In the end I went with the less committal word.
 * trying to avoid grammar syntax that may be hard to interpret by novices
    * for example the snippet `... ) '(' ( ...` is clear to an expert, but not immediately to a novice.
    * sadly, the trade-off for making individual rules simple is that we need more rules added in. David had me concede that the usability metric here shouldn't be "smallest possible grammar", so I tried to see if a couple of extra rules won't make the "applications" a little more obvious to read through.
    * With the PR, the grammar conveys directly that "an application is a function applied to a list of arguments". I think that may be a little better?

I also snuck in my preference for numbers being rigid (i.e. no leading or trailing decimal dots), since the alternative got added in a somewhat similar fashion. This ultimately depends on the resolution of #376 .

Can discuss and change as we see fit, but thought I would at least propose this perspective.

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -103,12 +103,15 @@ the user and agent.</p>
 
    <p>The <code class="attribute">intent</code>, after ignoring white
    space between tokens, should match the following grammar.</p>
-   
+
 <pre class="def bnf">
-intent   := numlit | NCName | argref | funcall
-funcall  := (NCName | argref | funcall) '(' ( intent (',' intent )* )? ')'
-numlit   := '-'? (digit+ '.'? digit* | '.' digit+)
-argref   := '$' NCName
+intent      := name | number | reference | application
+name        := NCName
+number      := '-'? digit+ ( '.' digit+ )?
+reference   := '$' NCName
+application := function '(' arguments? ')'
+function    := name | reference | application
+arguments   := intent ( ',' intent )*
 </pre>
 
 <p>Here <a href="https://www.w3.org/TR/REC-xml-names/#NT-NCName"><code>NCName</code></a>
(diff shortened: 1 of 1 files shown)


davidcarlisle (approved):
yes I think the grammar looks much better refactored this way, especially adding `arguments` allows an empty list just with `?` I wasn't too happy with the amount of complication this conceptually small change made to the previous presentation.

As you comment, this makes the separate change to restrict numeric literals. I'm less convinced by that, but since it is still discussed in an open issue, the current draft can have either form without greatly affecting the discussion, so I plan to merge as-is.

NSoiffer:
I like that better also. In fact, in my code, I already basically did that refactoring via a separate function when it needs to process the arguments. Great minds think alike. Probably feeble ones also :-)

PR #391, opened 2022-06-22
[intent grammar] variant without numeric literals

dginev:
Given the latest discussion in #376 , I believe I have lost what alignment I had with the group w.r.t numeric literals from the June 9th meeting.

As such, my only constructive suggestion is to avoid grammatically modeling numeric literals until a later date, which is what this PR enacts, simplifying the `application` rule respectively.

My understanding is that currently this is solely my own preference, so I don't expect a merge. Still, I wanted to make the suggestion clear.

In my own estimation, deferring numeric literals to a later release would still allow us to reap most of the benefit of having an intent attribute, and would give us time to properly collect and design for the cases where literal values are unavoidable.

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -105,17 +105,15 @@ the user and agent.</p>
    space between tokens, should match the following grammar.</p>
 
 <pre class="def bnf">
-intent      := name | number | reference | application
+intent      := name | reference | application
 name        := NCName
-number      := '-'? digit+ ( '.' digit+ )?
 reference   := '$' NCName
-application := function '(' arguments? ')'
-function    := name | reference | application
+application := intent '(' arguments? ')'
 arguments   := intent ( ',' intent )*
 </pre>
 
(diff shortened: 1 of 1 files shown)


brucemiller:
Can you succinctly remind me what excluding numeric literals accomplishes?

dginev:
succinctly: "low friction" in adoption

To reference [a slide](https://prodg.org/talks/encyclopedic-intent#/1/1) I presented to the group some time ago:

> 3. Low friction
>    - Good common questions: "what is this technical notation? how do we narrate it?"
>    - Bad common questions: "what is this intent syntax? how should I use it?"

brucemiller:
Interesting, but until we publish and analyse the response, we only have our instincts and experience to suggest whether certain features will lead to good or bad questions.  Personally, I expect that the lack of numeric literals would lead to more bad questions than their presence would.

dginev:
We found a workable temporary compromise in resolving #376 , so this PR will remain unneeded 👍🏻

NSoiffer:
To be clear, the decision occurred at the MathML meeting today.

PR #392, opened 2022-06-23
Fix mistakes

ghost:
I corrected some mistakes. Note that by the IEEE 754 standard, NaN is not equal to itself. Maybe you would like to replace the example “$0/0=\mathrm{NaN}$”.

### DIFF
--- a/index.html
+++ b/index.html
@@ -391,7 +391,7 @@ div.issue > *:nth-child(2) {display:block;}
     <section>
      <h4 id="parsing_rnc_core">MathML Core</h4>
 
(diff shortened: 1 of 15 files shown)


davidcarlisle (approved):
Thanks for this!
I have read all the diffs and all look good to me, but I'm travelling today and there is a WG meeting this evening I will miss, so approving but taking no action now, hopefully someone will merge after the meeting. On NaN yes we should make it say "is" or "returns" or some such rather than "="

davidcarlisle:
@NSoiffer  can I merge this now, or would you rather land your current chapter 3 edits first? There are loads of small diffs here so it would be good to merge it sooner rather than later while it has no conflicts...

NSoiffer:
My changes are only to maction and only involve throwing out most of the text with some new text that I'm still writing, so these changes won't be a problem for me.

PR #394, opened 2022-06-24
simplify intent grammar, dropping restriction on head of application

davidcarlisle:
The original subject of issue #376, the syntax for numeric literals, was decided in the WG call on 2022-06-23, however most of the recent discussion in that issue has been on whether to allow the head of a function call to be an unrestricted intent, or a restricted form not allowing a numeric literal.

@dginev  argued for keeping the restriction, I, @NSoiffer  and @brucemiller  argued (I think) for dropping the restriction and simplifyig the grammar to recursively call `intent` rather than have the extra `function` production which omits numeric literals.

To make sure we are all talking about the same change, this PR  makes explicit the suggested grammar

### DIFF
--- a/index.html
+++ b/index.html
@@ -206,6 +206,13 @@
                    "href": "https://w3c.github.io/mathml-docs/notes-on-mathml/"
   },
 			  
+  "MathML-AAM": {"title": "MathML Accessiblity API Mappings 1.0",
+                   "publisher": "W3C",
+                 "editors": ["Joanmarie Diggs", "Alexander Surkov", "Michael Cooper"],
+                   "status": "ED",
+                   "href": "https://w3c.github.io/mathml-aam/"
+  },
+			  
   "MathML-Media-Types": {"title": "MathML Media-type Declarations",
                    "publisher": "W3C",
                    "editors": ["Paul Libbrecht"],
--- a/src/accessibility.html
+++ b/src/accessibility.html
@@ -76,15 +76,13 @@
 
    <section>
     <h5 id="acc_aria">ARIA processing </h5>
-    <p>[Points to
-    MathML Accessibility API Mappings 1.0 [<a
-    href="https://w3c.github.io/mathml-aam/">https://w3c.github.io/mathml-aam/</a>]
-    &quot;This specification is intended for user agent developers
+    <p>Points to [[[MathML-AAM]]]
+    <q>This specification is intended for user agent developers
     responsible for MathML accessibility in their product. The goal of
     this specification is to maximize the accessibility of MathML
     content by ensuring each assistive technology receives MathML
     content with the roles, states, and properties it
-    expects.&quot;</p> <p>[Needs to be fleshed out...]</p>
+    expects.</q></p> <p>[Needs to be fleshed out...]</p>
    </section>
 
    
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -109,8 +109,7 @@ intent      := name | number | reference | application
 name        := NCName
 number      := '-'? digit+ ( '.' digit+ )?
 reference   := '$' NCName
-application := function '(' arguments? ')'
-function    := name | reference | application
+application := intent '(' arguments? ')'
 arguments   := intent ( ',' intent )*
 </pre>
 


NSoiffer (changes requested):
Although I think what you propose is simpler/better, we agreed at the meeting that we don't have a use case and the grammar should make it harder to do use a number as the head (it can still be done via `argref`).

brucemiller:
Hmm... I thought we had agreed to simplify as per @davidcarlisle's suggestion, but that we had agreed not to go overboard with exotic numbers until we had use cases.

dginev:
I remember what @NSoiffer remembers. Sounds like we may need another meeting to discuss?

davidcarlisle:
so we see the benefit of having an explicit diff to look at!

The meeting minutes only record discussion of number syntax, not any discussion of the function head (but I was not there,,,)

Keeping the restriction means more documentation and more tests will need to be added, and I see no benefit

davidcarlisle:
> I remember what @NSoiffer remembers. Sounds like we may need another meeting to discuss?

Since apparently this wasn't unambiguously decided at the last meeting I'm certainly happy to leave this PR open until the WG can decide in a future meeting. We probably can't all agree on which form is best, but at least we should have a clear minuted decision.

brucemiller:
Unless we are willing to add the complexity to exclude references to numbers (I'm not suggesting that), we shouldn't exclude them at all. The current grammar gives the misleading *impression* that numbers can't be used as a function.
The insufficiently devious implementer may not realize that they still can be. So, arguably, this restriction makes bugs more, not less, likely.

dginev:
> Unless we are willing to add the complexity to exclude references to numbers (I'm not suggesting that), we shouldn't exclude them at all. 

This restriction is too strong in my view. We have not committed to this level of defensive work to prevent forming grammatically invalid expressions via references until now. So why would the number case be special?

Here are some examples using references that are questionable/ungrammatical, do not involve numbers, yet are valid with the current design:

* empty nodes
  ```xml
  <mrow intent="$f($x,$y)">
    <mrow arg="f"></mrow>
    <mi arg="x">x</mi>
    <mspace arg="y" width="0px"></mspace>
  </mrow>
  ```

* unspecified defaults - glyphs
  ```xml
  <mrow intent="$op($x)">
    <mi arg="x">x</mi>
    <mo><mglyph arg="op" src="my-postfix-operator-glyph.png" width="20px" height="10px"></mglyph></mo>
  </mrow>
  ```

* unspecified defaults - text
  ```xml
  <mrow intent="$textop($x)">
    <mtext arg="textop">the successor, if any, of</mtext>
    <mspace width="10px"></mspace>
    <mi arg="x">x</mi>
  </mrow>
  ```

Note the use of punctuation in the mtext, which is *not* grammatical inside `intent` values.

So, to my eyes, the referencing machinery is not in any sense "well-typed" and the grammar is (currently) not a validation step when references are expanded.

---

The way I have understood the building blocks so far, I have been interpreting the expression `intent="$f($x)"` as if it meant:
- "narrate the subtree referenced by `$f` as if it were a function head applied to argument `$x`",

And I have not read it as:
- "the expression referenced by `$f` must precisely match the `function` rule in the intent grammar".

davidcarlisle:
@dginev 

>  Here are some examples using references that are questionable/ungrammatical, do not involve numbers, yet are valid with the current design:

There must be something missing in my understanding of your argument as they all seem to me to be examples of why we should not incur the costs of documenting (and enforcing with any intent checking tools) this restriction on the head of a function. 

As you show, many mathematically dubious expressions are allowed so it seems inconsistent to ban these.  Epecially when any system generating the readings has to be able to generate the same spoken forms if you have numeric heads via references. So it complicates the grammar and the description as we have to say where numbers may and may not be used, and has no practical benefit that I can see.  As far as I know, the existing mathcat implementation doesn't enforce this restriction.

dginev:
> As you show, many mathematically dubious expressions are allowed so it seems inconsistent to ban these.

That is (and has been) the case for the use of *references*.

It is not *yet* the case for the grammar syntax itself.

And I consider the grammar to have a bit of a self-documenting utility, where we have the opportunity to clearly communicate the **main** scope of intent to adopters.

I would say there is a "question of degree" that we need to settle on, essentially answering

"At what point does the design of `intent` get messy?"

At one extreme, I could see reasonable arguments for deferring to a "MathML-Next" the introduction of numbers, of higher-order function heads, or even of references, so as to avoid raising the adoption bar too high (what I call "low friction" in adoption).

On the other hand, and this is a point Bruce likes to champion, restricting the expressivity of intent enforces an aesthetic, and also smuggles in some assumed properties of how the presentation markup is structured (e.g. mrow scaffolding).

So, in that extreme, we may allow all ideas, and even avoid having a formalized grammar entirely, so as to, in your words:
> not incur the costs of documenting (and enforcing with any intent checking tools) this restriction

Clearly there ought to be a nice workable balance in the middle, where we get just enough expressivity that we have thought through, with as little confusion sprinkled in as possible.

And so, to conclude:
I don't find the expression `1(2)` to be on the admissible side, so I'm trying to avoid its inclusion. That comes in tandem with my growing fondness of having the extra clarity of well-named `function` and `application` rules in the grammar, which have a tempting level of suggestiveness to adopters.

davidcarlisle:
I think forcing extra implementation costs (or it seems implementations quietly ignoring the extra constraints) is a high price to pay for just some personal feeling that 1(1) falls on the wrong side of an arbitrary dividing line of wierdness. But yes I agree that writing specifcations is all about making such personal judgements. Just that in this case I don't think anyone else has pitched the line at the point you suggest. But as I said, I am happy to leave this open until after a meeting when the group can come to some sort of consensus.

brucemiller:
> We have not committed to this 
> level of defensive work to prevent forming grammatically invalid 
> expressions via references until now. So why would the number case be 
> special?

Thanks! That was exactly my point!

But you missed the other point: Intentionally complicating the grammar 
to make it look like a number can't be a function head, without actually 
achieving that goal would trick the unwary implementer on the one hand,
and induce skepticism otherwise. That's what *I* call High Anxiety.

NSoiffer:
A few points to back up various competing claims:
1. At the meeting, we did agree to dropping number heads from the grammar
2. For ease of implementation (not a big deal, but it avoids an extra param/check in my implementation), MathCAT does not enforce this restriction.

My feeling is that we are essentially arguing about how many angels fit on the head of a pin. Using a number head is weird and I don't know what anyone doing so would expect. I don't know of a use case and I don't think anyone will try to do it unless by mistake or malicious intent (trying to break something, akin to the many edge cases @dginev showed above). That's why I decided it wasn't worth arguing further and agreed with @dginev. Potentially disallowing it in the grammar might catch a mistake earlier if someone is using a validator, but there are so many ways of doing something wrong (e.g., doing the same via an argref), this is extremely minor -- authors should always verify that the expression looks and speaks correctly in at least on piece of software.

brucemiller:
>  1. At the meeting, we did agree to dropping number heads from the grammar

Maybe we did, but then I misunderstood what I was agreeing to.

davidcarlisle:
Quoting @NSoiffer 

> At the meeting, we did agree to dropping number heads from the grammar

the minutes did not record any discussion of function head, only of literal number syntax. (Not that that necessarily means no discussion took place)

> My feeling is that we are essentially arguing about how many angels fit on the head of a pin. 

It doesn't affect users much as they are unlikely to ever want to write 1(3) but it has a disproportionaely large effect on the spec and on any intent checking tools we want to produce as they have to document this restriction. It just feels very unnatural to try to use the grammar to constrain usage here as, as shown above, many other mathematically meaningless terms are allowed by the grammar.

But probably everyone has already said anything that can be said.  I propose an item for Thursday's meeting that this PR be merged. If the group decides against that, then so be it: I'll close this with no action, otherwise if they decide for, we can merge it.

dginev:
> many other mathematically meaningless terms are allowed by the grammar.

I will insist that they are allowed by the *references*, not by the *grammar*.

That distinction is likely the core difference in perspective we have at the moment.

davidcarlisle:
I am not sure why hiding weirdness via a reference makes any difference here but there are plenty of meaningless intents allowed without references  `rabbit(cat,dog)`  or `fraction(1,2,3)` the grammar is simply the wrong place to enforce semantic constraints

dginev:
Since this change was added in parallel through #399, this PR should probably be closed.

davidcarlisle:
overtaken by events, especially PR 399, closing with no action

PR #396, opened 2022-06-29
Deprecations

NSoiffer:
Deal with deprecations
* maxsize="infinity"
* longdiv & radical on menclose
* linebreak on mspace

Three notes:
1. the default value on menclose was changed to "box". Is this reasonable?
2. locally, I see a spurious ">" right before the table in http://localhost:8080/index.html#attributes-5. I can't see where that is coming from.
3. this picked up my previous PR https://github.com/w3c/mathml/commit/6a77210e27733f531516313d2a64291095de8107 -- github and I still don't see eye-to-eye


src/presentation-markup.html
```diff
@@ -2104,17 +2103,19 @@ <h6 id="presm_mo_dict_attrs">Dictionary-based attributes</h6>
...
  
        <tr>
         <td colspan="3" class="attdesc">
    Specifies the maximum size of the operator when stretchy;
    see <a href="#presm_op_stretch"></a>.
+   If not given, the maximum size is unbounded.
    <span>Unitless or percentage values indicate a multiple
-   of the reference size, being the size of the unstretched glyph.</span>
+   of the reference size, being the size of the unstretched glyph.
+   MathML 4 deprecates "infinity" as possible value as it is the same as not providing a value.</span>
```
davidcarlisle, line 2118: OK, but schema needs updating to match remove infinity from mathml4-presentation.rnc, add it to mathml4-legacy.rnc, action me, I guess

src/presentation-markup.html
```diff
@@ -5464,57 +5421,43 @@ <h5>Attributes</h5>
  
      <tr>
       <td rowspan="2" class="attname">notation</td>
-      <td>(<code class="attributevalue">longdiv</code> | <code class="attributevalue">actuarial</code> | <code class="attributevalue">phasorangle</code> | <code class="attributevalue">radical</code> | <code class="attributevalue">box</code> | <code class="attributevalue">roundedbox</code> | <code class="attributevalue">circle</code> |
+      <td>(<code class="attributevalue">actuarial</code> | <code class="attributevalue">phasorangle</code> | <code class="attributevalue">box</code> | <code class="attributevalue">roundedbox</code> | <code class="attributevalue">circle</code> |
```
davidcarlisle, line 5424: maybe empty would be a better default, so 

`<menclose><mi>x</mi></menclose>`

is same as 

`<menclose notation=""><mi>x</mi></menclose>`

and adding more words into the attributes always adds more decoration not first removing "box"

davidcarlisle (commented):
1. the default value on menclose was changed to "box". Is this reasonable?

   possibly not, see inline comment

2.    locally, I see a spurious ">" right before the table in http://localhost:8080/index.html#attributes-5. I can't see where that is coming from.

      There was a spurious `>` before `</tbody>` also a duplicated `</table>`  I fixed both in this branch in  7def6d2 (The Html parser "corrects" text within  a table but not in a row by moving it before the table so when you inspect the spurious text it doesn't show an error and doesn't show where it was before the parser moved it, it simply shows as text  before the table. )

3.  this picked up my previous PR https://github.com/w3c/mathml/commit/6a77210e27733f531516313d2a64291095de8107 -- github and I still don't see eye-to-eye
  

    It would have been simpler to branch off main again (without the maction changes) but doesn't really matter.  Note that github PR show a three dot diff showing all changes since a common ancestor, you can see the (typically smaller) diff between the branch and the head of main using a two dot diff. Compare:


Three: https://github.com/w3c/mathml/compare/main...deprecations

Two: https://github.com/w3c/mathml/compare/main..deprecations


leaving open for  now pending decision on menclose box. Otherwise all looks good

src/presentation-markup.html
NSoiffer: I added text about menclose with no notation attr (or an empty string) should do nothing.

Is there anything else to resolve?

Issue #398, opened 2022-06-30
stuck with functional speech?


NSoiffer:
This issue pulls the essence of the question I raised about chemistry in #397 and which is also part of prefix/infix/postfix speech mentioned in #256. This issue only applies to non-core intents. In #397, the proposed solution is to move the chemistry notations into core. This doesn't solve the underlying problem I was trying to point out in that issue.

On the call today, I think it was @davidfarmer who brought up another example. It may have been p-norms or something else. I'm going to stick with p-norms as an example because they point out a problem. The notation is $||\\mathbf{x}||_p$ with a MathML encoding:
```
<msub>
   <mi mathvariant='bold'>x</mi>
   <mi>p</mi>
</sub>
```
A very reasonable `intent` value would be `intent='p-norm($base, $p)'` where the `p` and the `x` in the above example are marked appropriately. Being in "open", this would likely get a verbal rendering something like "p norm of x and p". Not ideal, so maybe one would write  `intent='p-norm($base)'` to get "p norm of x".

But what if the subscript is `4`? I think in that case you want to hear "4 norm of x". Maybe that's not the standard speech for this construct, but for another notation, it might be desired if it isn't here. So for the sake of discussion, assume that it is the desired form to pronounce this notation.  That means the subscript should either be one of the arguments or be part of the name of the function. If the subscript needs to part of the name of the function, that means that one can't put into the open list "p-norm" and be done with it, one needs to say that there are an infinite number of names `2-norm`, `3-norm`, etc. I don't think that notion currently exists in the open list. Note that those names (`2-norm`, etc.) aren't actually valid with the current syntax since you can't start a name with a digit, so `two-norm`, `three-norm`, etc. 

Clearly there is a parameter here and the problem is that although `intent` is meant for speech, `intent` forces a single style of speech for notations in open. I don't want to make `intent` have a complicated syntax, but I also want `intent` to be able convey what the author wants to be said for a given notation. In this case, it is a more infix style.  This _could_ be conveyed as `silent-intent="p-norm($p, "norm-of", $x)"`. Another option _could_ be `infix-intent="norm-of($p, $x)"` where the function head is intended to be used between the args. A `silent-intent` would have solved the chemistry speech issue in #397. In #256, I also proposed `prefix-intent` and `postfix-intent`, but with `silent-intent`, there is no need for them. 

[...]

Issue #337, opened 2022-05-06
Intent for "mtable" constructs


dginev:
We have discussed some early variations of intent annotations for tabular constructs, thanks to Sam's [prototype document](https://mathml-refresh.github.io/discussion-papers/intent#tables).

I don't think we have a place to finish the discussion yet, similarly to the issue we made for n-ary operations in #253 .

There are at least three questions here:
 * What do current AT tools struggle with in the tabular MathML examples we have, i.e. where do we need annotation beyond the baseline narration.
    * Preliminary: gather+systematize enough tabular examples.
 * What would be an acceptable well-structured narration for each case?
    * Is there a default navigation pattern? 
    * How does one prefer to navigate unusual walks in a tabular, from AT's perspective ?
 * When we know what AT aims to produce, we can figure out appropriate **values** to use in `intent`, to guide the tabular walk.

---

Tonight I found a good non-standard example to add in: a MathML diagram. Such constructs are common in research documents that end up in e.g. arXiv.

The diagram can be found in MathML at [ar5iv:2204.13136](https://ar5iv.labs.arxiv.org/html/2204.13136#S1.Ex1.m1.3) and renders in Firefox as:

![On robustness and related properties on toric ideals](https://user-images.githubusercontent.com/348975/167084343-10656af9-7a05-43fe-bc31-66dd99b8b0a9.png)

The latexml-generated MathML is the following (you need to expand the details tag):
<details>
<summary>Details</summary>

```html
<mtable columnspacing="5pt" rowspacing="0pt" displaystyle="true">
  <mtr>
    <mtd columnalign="center">
      <mrow>
        <mi class="ltx_font_mathcaligraphic">𝒮</mi><mo>=</mo>
        <mrow>
          <mo stretchy="false">⟨</mo>
          <msub><mi>a</mi><mn >1</mn></msub>
          <mo>,</mo>
          <msub><mi>a</mi><mn >2</mn></msub>
          <mo stretchy="false">⟩</mo>
        </mrow>
      </mrow>
    </mtd>
    <mtd columnalign="center"><mo>⟺</mo></mtd>
    <mtd columnalign="left">
      <mrow><mi class="ltx_font_mathcaligraphic">𝒮</mi><mo>⁢</mo>
        <mtext>&nbsp;is robust&nbsp;</mtext>
      </mrow>
    </mtd>
  </mtr>
  <mtr>
    <mtd columnalign="center"><mo>⇓</mo></mtd>
    <mtd></mtd>
[...]

PR #399, opened 2022-07-09
Davidc/intent hint

davidcarlisle:
Discussions about intent are still ongoing but it seems a draft based on the issue comment at

https://github.com/w3c/mathml/issues/398#issuecomment-1178880279

Is closer to a final consensus position than the current draft.

I doubt if any wording here is final but I'd propose we merge this to form a new baseline against which people can make comments and new issues.

It would be good to merge this before Thursday's meeting so the gh-pages view is easily seen.

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -1,7 +1,7 @@
 <section>
  <h2 id="mixing">Annotating MathML</h2>
-<!-- <div class="issue" data-number="281"></div>-->
-<!-- <div class="issue" data-number="252"></div>-->
+ <!-- <div class="issue" data-number="281"></div>-->
+ <!-- <div class="issue" data-number="252"></div>-->
 
  <div class="issue" data-number="289"></div>
  <div class="issue" data-number="257"></div>
@@ -25,11 +25,10 @@
 
  <section>
   <h3 id="mixing_intent">The <code class="attribute">intent</code> attribute</h3>
-  <p>As decribed in <a href="#presm_presatt_intent"></a>, Presentation
-  MathML elements
-  allow attributes <code
+  <p>As described in <a href="#presm_presatt_intent"></a>, Presentation
+  MathML elements allow attributes <code
   class="attribute">intent</code> and <code
-  class="attribute">arg</code> that as described here allow the
+  class="attribute">arg</code> that allow the
   <dfn><q>intent</q></dfn> of the term to be specified. This
   annotation is not intended to provide a full mathematical definition
   of the term but may be used to guide accessibility systems to give audio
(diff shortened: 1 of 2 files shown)


NSoiffer (commented):
There's a lot of text -- any chance you can provide a preview of this branch on your github-pages?

davidcarlisle:
> There's a lot of text -- any chance you can provide a preview of this branch on your github-pages?

@NSoiffer  You can always view locally, but anyway, a few pulls and pushes , and I synced my fork so

https://davidcarlisle.github.io/mathml/#mixing_intent

NSoiffer:
You should add a link for 398 to the start of the section.  I hope we can at least close some of the other discussions/notes.

"Intent Concept Dictionary" -- this is undefined (and the first I've heard of it).

> A literal is intended to correspond to some mathematical concept or operator

I think you mean "concept"

> Note that [intent](https://davidcarlisle.github.io/mathml/#dfn-intent) identifiers are maintained in two lists, currently located at ...

You can drop "Note that", or maybe move the whole paragraph into an issue link or editor's note. 

>In a function [application](https://davidcarlisle.github.io/mathml/#intent_application), if the function is followed by a [hint](https://davidcarlisle.github.io/mathml/#intent_hint) the reading of the name may be suppressed,

missing a comma after "hint" (so actually after the close paren of the link).

I still think munging the examples together with "->" in between is mathematical nonsense, something we have avoided. The complex conjugate is better, but it is still mathematically "silly" (which is a well-defined math term :-)

The audio renditions contain "_"; they should be translated to words ("bar").

The examples of "hint" should be "real" ones:
prefix: !x (as in "not x")
infix: (×) cross-product
postfix: double-factorial
silent:  negative-ion -- Na+  [that really needs renaming]
Not all of these are ideal because putting the name on the operator would do the right thing -- one doesn't need `intent` on a parent `mrow`.

> `<mtr intent="append">`

This probably should be `append@silent`, although I don't think "append" is a great word. Among other things, it is not in the "conceptual dictionary". Using that dictionary, the `mtable` should have "system-of-equations" and the `mtr`s should have "aligned-equation". Being in core, they probably don't need "@silent".

I'm not sure what the appropriate "intent" for the second one is. @dginev?

dginev:
> I'm not sure what the appropriate "intent" for the second one is. @dginev?

This PR is likely not the best place to solve #337 . Or a new issue for "Tabular Intent Core", if you think that would be helpful. As much as I understand the pain of having so much message volume on github, inventing all of this infrastructure requires it.

davidcarlisle:
> > I'm not sure what the appropriate "intent" for the second one is. @dginev?
> 
> This PR is likely not the best place to solve #337 . Or a new issue for "Tabular Intent Core", if you think that would be helpful. As much as I understand the pain of having so much message volume on github, inventing all of this infrastructure requires it.

I'd agree a new issue (or just re-starting that issue) would be better than discussing individual examples here.

But one reason there has been little progress on intent for specific constructs, is failure to tie down a basic syntax, which makes it hard to discuss anything explicit. Which is why I think we need to merge something like this PR sooner rather than later, so the other issues can be discussed using a syntax we can use at least for the first working draft

davidcarlisle:
quoting @NSoiffer 

> You should add a link for 398 to the start of the section. I hope we can at least close some of the other discussions/notes.
> 


**done**  (in the branch, I suppose I'll re-sync the fork later, grmble)

> "Intent Concept Dictionary" -- this is undefined (and the first I've heard of it).
> 

Trying out a new name for "intent lists"  Inspired mostly by Deyan's pushing for clearer separation between NCName being an entry to a known concept and the use as a Literal Name read directly.  "list" sounds like a list of names, "Dictionary" giving an idea of the muliple fields, but open to suggestions on what to call  the things "open" and "Core"

> > A literal is intended to correspond to some mathematical concept or operator
> 
> I think you mean "concept"
> 
Yes probably (That phrase came from an old CG document, I'll adjust)

> > Note that [intent](https://davidcarlisle.github.io/mathml/#dfn-intent) identifiers are maintained in two lists, currently located at ...
> 
> You can drop "Note that", or maybe move the whole paragraph into an issue link or editor's note.
> 

I dropped Note that. This is still calling them "lists" we should do a full sweep over the doc once we decide what to call them. (Perhaps list is OK....)


> > In a function [application](https://davidcarlisle.github.io/mathml/#intent_application), if the function is followed by a [hint](https://davidcarlisle.github.io/mathml/#intent_hint) the reading of the name may be suppressed,
> 
> missing a comma after "hint" (so actually after the close paren of the link).

**done** 
> 
> I still think munging the examples together with "->" in between is mathematical nonsense, something we have avoided. The complex conjugate is better, but it is still mathematically "silly" (which is a well-defined math term :-)
> 

Yes I know. Have I mentioned not agreeing a syntax has been a dis-incentive to generating better examples:-) 

Probably once we have more real examples highlightig specific features: core v open, hints or not, tables, ... Then that block of initial generic examples can go. So I am not too worried about the arrows at this point. (Although I could for now just split it up into multiple examples?)


> The audio renditions contain "_"; they should be translated to words ("bar").
> 
> The examples of "hint" should be "real" ones: prefix: !x (as in "not x") infix: (×) cross-product postfix: double-factorial silent: negative-ion -- Na+ [that really needs renaming] Not all of these are ideal because putting the name on the operator would do the right thing -- one doesn't need `intent` on a parent `mrow`.
> 

Also I wasn't sure if it was clearer to give an example of just the intent or a mathml fragment so

`inner-product @infix ($x,$y)`

or

```
<mrow intent="inner-product @infix ($x,$y)">
  <mi arg="x">A</mi>
  <mi arg="y">B</mi>
</mrow>
```

> > `<mtr intent="append">`
> 
> This probably should be `append@silent`, although I don't think "append" is a great word. Among other things, it is not in the "conceptual dictionary". Using that dictionary, the `mtable` should have "system-of-equations" and the `mtr`s should have "aligned-equation". Being in core, they probably don't need "@silent".

that whole section just comes from some early CG document. If we do specify something for table rows I'd assume it would be core and not need any hint

> 
> I'm not sure what the appropriate "intent" for the second one is. @dginev?

Post merge issues:-)

davidcarlisle:
@NSoiffer  arrow example split up https://davidcarlisle.github.io/mathml/#mixing_intent_examples_notation
cf82849

brucemiller (approved):
This is a helpful step forward to focus further discussion, refinements and (possibly) retractions.

NSoiffer:
Thanks for the changes. Clearly this is not the final version, so it doesn't have to perfect. My suggestions were just to fix some typos and clarify some things.

PR #400, opened 2022-07-14
more text on silent hint and underscore

davidcarlisle:
Possible text as discussed in #398  warning about possible downsides of forcing text and describing a (weak form of) the `_` prefix just as a posible naming  convention you might want to follow.

HTML view at

https://davidcarlisle.github.io/mathml/#mixing_intent_examples_hint

(only that section changed)

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -298,13 +298,42 @@
 
      <li><code>semi-factorial @postfix($x)</code> would be read as <q>x semi factorial</q></li>
 
-     <li><code>free-algebra ($r, $x)</code> would be read as <q>free algebra of r and x</q></li>
-     
-     <li><code>free-algebra-construct@silent (free, $r, algebra-on, $x)</code> would be read as <q>free, r algebra on x</q></li>
-
-
     </ul>
 
+    <p>In most cases it is preferable to use an <code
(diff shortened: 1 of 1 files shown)


brucemiller:
If we're adopting the convention that "_" prefixed names are treated as literals, that probably deserves mention in the `concept-or-lteral` description above. Perhaps just modify the 2nd bullet to start:
```
If the name does not match a Concept Name known to the application or begins with _ (U+005F), a default reading is 
```
Eventually, we'll want to assert that such names will not/should not be added to either the core or open dictionaries.

Even if we can't enforce this convention, it seems useful. Eventually somebody's going to use "power" where they should have used "_power" and get some weird speech. :>

davidcarlisle:
@brucemiller  I adjusted the wording 0ef2619bd31a9ad11bb723bfb4f83fa2dad1331b

brucemiller (approved):
I'm still hoping there's a better way, but this nicely explains the current proposal.

davidcarlisle:
@brucemiller  yes made a PR mostly as a way to get an easy interface to a diff for suggested text as far as I understood the proposal, not necssarily to merge. Although we could merge it if that ends up being the  agreed plan.

PR #401, opened 2022-07-14
Added text to the subsection "Accessibility benefits of using MathML"

Steve-Noble:
It is my understanding that I should add updates as often as possible, so I added some text I have so far. Will need to go back later to add some references, and perhaps subheadings, but the words are essentially all there as I want them for this subsection.


davidcarlisle (commented):
As you commented in the meeting the whole appedix will need further review once the words all there but this seems a nice incremental improvement filling out this section

Issue #402, opened 2022-07-15
sibling intent references


davidcarlisle:
Currently we say

> references prefixed with `$`. An argument reference such as `$name` references a descendent element that has an attribute `arg="name"`. Unlike id attributes, arg do not have to be unique within a document. When searching for a matching element the search should only consider descendants, while stopping early at any elements that have a set intent or arg attribute, without descending into them.

We could say

> references prefixed with `$`. An argument reference such as `$name` references a descendent **or sibling** element that has an attribute `arg="name"`. Unlike id attributes, arg do not have to be unique within a document. When searching for a matching element the search should **first** consider descendants, while stopping early at any elements that have a set intent or arg attribute, without descending into them. **If this search fails, and the current element is not `<math>`, the search is repeated from the parent element.**

---

This would allow intents on groups of table rows:

```
<mtr intent="sum($a,$b,$c)"><mtd><mi arg="a">a</mi></mtd><mtd><mo>+</mo><mi arg="b">b</mi></mtd></mtr>
<mtr>                       <mtd></mtd>                  <mtd><mo>+</mo><mi arg="c">c</mi></mtd></mtr>
```

It would also allow (which may not be seen as an advantage) expressions missing nested mrow structure to be annotated with `intent` without having to add in the missing mrow.

1 + log x

```
<mrow>
<mn>1</mn>
<mo>+</mo>
<mi arg="l" intent="$l($x)">log</mi>
<mi arg="x">x</mi>
</mrow>
```

dginev:
I have a kind of a pedantic point, but to my reading:
```xml
<mi arg="l" intent="$l($x)">log</mi>
```

should expand into an infinite `log(log(log(...log(x)))...)`. The self-reference is generally a scary proposal.

---

If the descendant-only `arg`s are getting to be too constraining, why not allow an additional `#id` reference, for the rare cases where one has to express such "far away" arguments:

```xml
<mtr intent="plus($a,$b,#c)">
  <mtd><mi arg="a">a</mi></mtd>
  <mtd><mo>+</mo><mi arg="b">b</mi></mtd>
</mtr>
<mtr>
  <mtd></mtd>
  <mtd><mo>+</mo><mi id="c">c</mi></mtd>
</mtr>
```

I admit this is biased, as a common latexml workflow is to already generate id attributes.

Personally, I'm fine without *either* extension, and imposing a constraint that keeps some simplicity in reasoning about the expressions.

[...]

PR #403, opened 2022-07-21
ednote for mtable intents

davidcarlisle:
The current draft and discussions on `intent` for tables in #337 and #402 don't align. I am sure it is fixable and we can propose a workable solution, but I'm not sure  we can do it for the FPWG, so this PR removes the intent=... attributes from the mtable examples in 5.1.2.4 and adds an Editor's Note saying we will address this next time.

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -379,20 +379,39 @@
 
 
    <section>
-    <h4 id="mixing_intent_examples_mtr">Tables</h4>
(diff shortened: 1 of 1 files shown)


dginev (approved):

brucemiller (approved):
I've been watching the discussion on table intent; a bit queasy, but no constructive comments yet. 

But, yeah, we should procrastinate a bit longer.

PR #406, opened 2022-07-28
Moved rewrite:element from 4.3.4 to Appendix F

### LINKED ISSUE #356: rewrite: element transformation
**rewrite: element**  is at https://w3c.github.io/mathml/spec.html#dfn-rewrite-element is in section 4.3.4 and the only rewrite rule still in Chapter 4. I wonder if it should move to appendix G? It would need some minor rewordings of the adjacent paragraphs.

samdooley:
Closes #356

### DIFF
--- a/src/contm-new.html
+++ b/src/contm-new.html
@@ -2841,53 +2841,24 @@
    <section>
     <h4 id="contm_opclasses">Operator Classes</h4>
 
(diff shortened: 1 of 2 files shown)


samdooley:
https://samdooley.github.io/mathml/#contm_opclasses

PR #408, opened 2022-08-04
alignment points

NSoiffer:
Big changes to maligngroup/malignmark to eliminate their attributes and simplify the description.

Also, malignmark can no longer occur at any point, including inside tokens.

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -7387,10 +7387,18 @@
    <section>
     <h5>Deprecation Notice</h5>
     <p>
-      With one significant exception, <code class="emptytag">&lt;maligngroup/&gt;</code> and <code class="emptytag">&lt;malignmark/&gt;</code> have had minimal adoption
-      and implementation. The one exception only uses the basics of alignment. Because of this, alignment in MathML is significantly
-      simplified to align with the current usage and make future implemntation simplier. In particular, all of the attributes
-      for these elements have been deprecated.
+      With one significant exception, <code class="emptytag">&lt;maligngroup/&gt;</code> and <code class="emptytag">&lt;malignmark/&gt;</code>
+      have had minimal adoption and implementation.
+      The one exception only uses the basics of alignment. Because of this, alignment in MathML is significantly
+      simplified to align with the current usage and make future implemntation simplier. In particular, the following simplifications are made:
+      <ul>
+        <li>the attributes for <code class="emptytag">&lt;maligngroup/&gt;</code> and <code class="emptytag">&lt;malignmark/&gt;</code> have been deprecated</li>
+        <li>The <code class="attribute">groupalign</code> attribute previously allowed on
+          <code class="element">mtable</code>, <code class="element">mtr</code>, and 
+          <code class="element">mtr</code> is deprecated</li>
+        <li><code class="emptytag">&lt;malignmark/&gt;</code> used to be allowed anywhere, including inside of token elements;
+          it is now allowed in only the locations that <code class="emptytag">&lt;maligngroup/&gt;</code> is allowed (see below)</li>
+      </ul>
     </p>
    </section>
    <section>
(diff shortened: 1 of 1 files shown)


src/presentation-markup.html
```diff
@@ -7387,6 +7384,23 @@ <h5 id="presm_mtdatts">Attributes</h5>
    <h4 id="presm_malign">Alignment Markers
    <code class="emptytag">&lt;maligngroup/&gt;</code>, <code class="emptytag">&lt;malignmark/&gt;</code><span class="coreno"></span></h4>
  
+   <section>
+    <h5>Deprecation Notice</h5>
+    <p>
+      With one significant exception, <code class="emptytag">&lt;maligngroup/&gt;</code> and <code class="emptytag">&lt;malignmark/&gt;</code>
+      have had minimal adoption and implementation.
+      The one exception only uses the basics of alignment. Because of this, alignment in MathML is significantly
+      simplified to align with the current usage and make future implemntation simplier. In particular, the following simplifications are made:
```
davidcarlisle: implementation simpler

src/presentation-markup.html
```diff
@@ -7482,61 +7483,49 @@ <h5>Description</h5>
...
-    <p>The alignment groups start at the locations of invisible
-    <code class="element">maligngroup</code> elements, which are rendered with
-    zero width when they occur outside of an alignment scope, but within
-    an alignment scope are rendered with just enough horizontal space to
-    cause the desired alignment of the alignment group that follows
-    them. A simple algorithm by which a MathML application can achieve this is given
-    later. In the example above, each equation would have one
+    <h5 id="presm_specify_align_group">Specifying alignment groups</h5>
+    <p>Each part of expression to be aligned should be in an <code class="element">maligngroup</code>.
+       The point of alignment is the left edge (right edge for RTL) of the element that follows an <code class="element">maligngroup</code> element
```
davidcarlisle: (right edge if `<code>dir="rtl"</code>`)

src/presentation-markup.html
```diff
@@ -7578,55 +7567,40 @@ <h5>Specifying alignment groups</h5>
...
     alignment.</p>
  
-    <p>Note that some positions for an <code class="element">maligngroup</code>
-    element, although legal, are not useful, such as  an
-    <code class="element">maligngroup</code> element <span>that is</span> an argument of an
-    <code class="element">mfenced</code> element. <span>Similarly, when inserting an <code class="element">maligngroup</code>
-    element in an element whose arguments have positional significance, it
-    is necessary to introduce a new <code class="element">mrow</code> element
-    containing just the <code class="element">maligngroup</code> element and the child element
-    it precedes in order to preserve the proper expression structure.  For
```
davidcarlisle, line 7588: we could make `<mfrac><mn>1</mn><malignmark/></mfrac>` schema invalid,  previously they were valid as malignmark and maligngroup listed in PresentationExpression so allowed anywhere but they could be treated like `<none/>` and only allowed in specific contexts. The mathml4-legacy schema can make them valid again (and add attributes back, etc...

src/presentation-markup.html
```diff
@@ -7691,382 +7665,29 @@ <h5>Table cells that are not divided into alignment groups</h5>
...
-    <p>If there is more than one <code class="element">malignmark</code> element
+    alignment group, it should only occur within the elements allowed for <code class="element">maligngroup</code>
+    (see <a href="#presm_specify_align_group"></a>).
+    If there is more than one <code class="element">malignmark</code> element
     in an alignment group, all but the first one will be ignored. MathML
     applications may wish to provide a mode in which they will warn about
     this situation, but it is not an error, and should trigger no warnings
     by default. <span>The rationale for this is that</span> it would
     be inconvenient to have to remove all
     unnecessary <code class="element">malignmark</code> elements from
```
davidcarlisle, line 7651: Why not make it an error?

davidcarlisle (commented):
I suspect we should remove the features rather than deprecate.

Currently there is a broken link as `groupalign` is described linking to `#group-alignment-list-list` which no longer exists, but more generally I think "deprecated" means "if you use it, it should work, but the feature may be removed in future" So the benefit of simpler schema and simpler text not gained until MathML5.

So "deprecation" would mean we still have to describe the old full version, but then advise it isn't used.

I think we should treat this as if it had been deprecated in MathML3,  so that `groupalign=` or `<mn><maligmark/></mn>` can be marked invalid in mathml4 unless using `mathml4-legacy` schema.

davidcarlisle:
@NSoiffer 's branch is visible at

https://davidcarlisle.github.io/mathml/#deprecation-notice

src/presentation-markup.html
NSoiffer: fixed

src/presentation-markup.html
NSoiffer: fixed

src/presentation-markup.html
NSoiffer: It is more consistent with the "malignmark/maligngroup" can occur anywhere philosophy (ignore if not appropriate) to ignore them.

src/presentation-markup.html
NSoiffer: Sounds reasonable.

I won't mark this as resolved since you have some work to do.

src/presentation-markup.html
davidcarlisle: OK I suppose (personally I'd make them invalid, making it more likely editors avoid this, browsers silently ignore invalid markup in any case) But not a battle I want to fight so I'll go with the flow here, marking resolved

davidcarlisle (approved):
schema is (still) in mathml-refresh github so I think all done here, I'll merge. I hope to update schema before the call this (UK) evening

PR #411, opened 2022-08-11
Intent

NSoiffer:
Rewrote introduction to Annotating MathML so that both semantics and intent are hopefully introduced more appropriately.

Second commit rewrote (amplified) some of the text about "intent" and specifically the core/open explanations.


davidcarlisle (approved):
I made one minor change, 5ca1710 otherwis looks good, merging.

PR #412, opened 2022-08-11
Rewording and reorganizing introduction to fit with the current draft

brucemiller:
As it says...
Should be closer to the final form.

### DIFF
--- a/src/introduction.html
+++ b/src/introduction.html
@@ -15,10 +15,6 @@
   of the mathematics being represented, the notation and semantics
   are not the same.  The semantic symbols and structures are
   subtly distinct from those of the notation.</p>
- </section>
-
- <section>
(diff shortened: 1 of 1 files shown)


src/introduction.html
```diff
@@ -44,43 +40,58 @@ <h3 id="intro_overview">Overview</h3>
...
   </p>
- </section>
 
+
+
+ </section>
+ <section>
+  <h3 id="intro_overview">MathML Notes</h3>
+  <p>
+   In addition to these two specifications, the Math WG group has developed the non-normative
```
davidcarlisle, line 92: we could also mention media types spec at this level  (and probably in the end an update to the entities spec) but I don't think this comment blocks this PR

davidcarlisle (approved):

PR #413, opened 2022-08-13
pubrules security and privacy sections

davidcarlisle:
Changing the class from ED to FPWD triggered some extra checks from respec and pubrules, and both wanted security and privacy considerations sections.

This PR adds appendices with the required names with initial text just referencing the matching sections in mathml-core. We could add more specific text if required later, but opening a PR to log a decision to add or not these sections as appendices.

### DIFF
--- a/index.html
+++ b/index.html
@@ -193,7 +193,10 @@
      edDraftURI:    "https://w3c.github.io/mathml/",
      shortName:     "MathML4",
      canonicalURI: "TR",
-     maxTocLevel:   4,
+       maxTocLevel:   4,
+       lint: {
+	   "required-sections":false
+       },
      otherLinks:    [{
      key: "Latest MathML Recommendation",
      data: [{
@@ -485,8 +488,6 @@ div.issue > *:nth-child(2) {display:block;}
   <section data-include="src/operator-dict.html" data-include-replace="true"></section>
   <section data-include="src/accessibility.html" data-include-replace="true"></section>
   <section data-include="src/conformance.html" data-include-replace="true"></section>
-  <section data-include="src/security.html" data-include-replace="true"></section>
-  <section data-include="src/privacy.html" data-include-replace="true"></section>
   <section data-include="src/contm-ops.html" data-include-replace="true"></section>
   <section data-include="src/transform2strict.html" data-include-replace="true"></section>
   <section data-include="src/mmlindex.html" data-include-replace="true"></section>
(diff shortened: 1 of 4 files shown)


davidcarlisle:
We need these sections but they don't have to be top level appendices so the same markup could instead be added at the end of  `conformance` so making sections D.4 and D.5 instead of E and F. So keeping the total number of appendices down

NSoiffer:
I looked a few specs and they all had a single section labelled "Privacy and Security Considerations". Some were small and some larger. Most started with "This section is non-normative."  I'd suggest doing that except that the rest of the spec is sloppy about it, so I don't know if we should clean up the rest of the spec and put it in here or just ignore this.

FYI: ARIA for example did what you wrote and punted to HTML for security and privacy.

I guess my only actionable suggestion is to combine the two sections into one.

polx:
I think something around here should speak about the DTD, schema or entity injection danger... At least MathML core does not seem to do so.

davidcarlisle:
@polx  can you suggest some text? I am not sure what you have in mind here as MathML doesn't control the host document, and as the only normative schema is the relaxNG which can not be referenced from the document, I'm not sure what we could say,

NSoiffer:
I'm good with what is there. Once @polx signs off or suggests some text [seems not needed to be, but not my area], go ahead with the merge.

polx:
Something such as the following?

In some situations, MathML expressions can be parsed as XML. The security considerations of XML parsing apply then as explained in [RFC 7303](https://www.rfc-editor.org/rfc/rfc7303#section-10).

polx:
All good.
I just don't see why [?RFC has a ? but it seems to not bother.

davidcarlisle:
@polx  ?`makes it non-normative although actually it make no difference here as references in non normative appendices are automatically non-normative. 

so this added RFC7303 to J.2 although I notice J.1 already has RFC3023 with same title,  I guess we should just reference 7303 ?

3023 refs:

```
$ grep 3023 index.html src/*l
src/media-types.html:    application/xml as specified in [[RFC3023]]</p>
src/media-types.html:    [[RFC3023]] apply.</p>
src/media-types.html:    <p>As with other XML types and as noted in [[RFC3023]] section 10,
src/media-types.html:    <p>Magic number(s): see [[RFC3023]]</p>
src/media-types.html:    application/xml as specified in [[RFC3023]]</p>
src/media-types.html:    [[RFC3023]] apply.</p>
src/media-types.html:    <p>As with other XML types and as noted in [[RFC3023]] section 10,
src/media-types.html:    <p>Magic number(s): see [[RFC3023]]</p>
src/media-types.html:    application/xml as specified in [[RFC3023]]</p>
src/media-types.html:    [[RFC3023]] apply.</p>
src/media-types.html:    <p>As with other XML types and as noted in [[RFC3023]] section 10,
src/media-types.html:    <p>Magic number(s): see [[RFC3023]]</p>
src/world-interactions.html:   which has been superseded by [[RFC3023]].</p>
```

polx:
well, that file could be removed. I've created the pull request [#414](https://github.com/w3c/mathml/pull/414).

Issue #306, opened 2022-03-31
extract media-type declarations into a standalone specification


polx:
The text in chapter 6 about the resource-types and encodings (e.g. [6.2.3](https://w3c.github.io/mathml/#world_resource_types) and [6.2.4](https://w3c.github.io/mathml/#encoding-names) and the text in [appendix B](https://w3c.github.io/mathml/#media-types-reg) should be extracted into a note that will enter [mathml-docs](https://github.com/w3c/mathml-docs/) and should be prepared to become a recommendation.

polx:
First effort made in [this branch](https://github.com/w3c/mathml-docs/tree/extract-media-type-issue-306).

@davidcarlisle : is there a way for me to have more "wrappings" such as table of contents and a bibliography?
And... how does it get rendered? (I guess into [here](https://w3c.github.io/mathml-docs/mathml-media-types/)?).

Tricky parts thus far:

- refer to MathML's chapter 3 and chapter 4 as the definition of presentation and content is done using such sa https://www.w3.org/TR/MathML/chapter3.html . I guess it is stable enough.
- refer to MathML's declaration is done for MathML3. I think that this stays.

Future: once the [RFC6388](https://datatracker.ietf.org/doc/html/rfc6838) is moved to drop Macintosh 4-letter codes, we shall do so...

davidcarlisle:
@polx to see a rendered version locally I just use a web server rooted at the git checkout, specifically

`python3 -m http.server 8080 &`

then look at ` http://localhost:8080/index.html`  

to get a public view it's easier to do as @samdooley  just did for the chapter 4 re-arrangment, use a fork and run gh-pages on the fork ending up in that caase with  https://github.com/samdooley/mathml

You should get a table of contents automatically just from  using correctly nested `<section><h???>xxxx` in the main document.

For references reference a spec such as `[[MathML3]]` and respec will make a references section. respec knows about all w3c and rfc and similar specs. It doesn't know about MathML4 yet, you can add more though , see the `localBiblio`  entry in the respec config of the main spec. You can check what resepc does already know about at https://www.specref.org/

davidcarlisle:
@polx  the table of contents didn't appear as there was a respec error (the mmlindex function copied from the main spec had no target)  I have fixed it up locally and some other respec warnings.

 I could push to your branch if you have no pending edits or I could make a new branch and PR to your branch if you prefer?

[...]

PR #414, opened 2022-08-13
Media types have been extracted

polx:
Following [this comment of @davidcarlisle](https://github.com/w3c/mathml/pull/413#issuecomment-1214211155): removing media-types (they have been published as part of issue #306) and pointing to RFC7303 instead of the older one.


davidcarlisle:
@polx  thanks the change looks good, but spec.html shouldn't be in main branch (just gh-pages)  I'll apply the change locally  then close here.

davidcarlisle:
closing here as done

PR #415, opened 2022-08-18
Intent

NSoiffer:
I hope I got this right... I probably should have deleted the branch first and the recreated it, but I did do a rebase. It looks like @davidcarlisle did a commit between the time of the rebase and my commit :-(.

What I did:
More explanation for intent.
Based on discussions at the Aug 11 meeting, I added the hint value "function" to account for the difference between prefix and functional forms.


davidcarlisle:
The PR shows 30 commits over 15 files so  a bit hard to see the wood for the trees, However the two dot compare shows all changes are in chapter 5:


https://github.com/w3c/mathml/compare/main..intent

NSoiffer:
The only changes are in src/mixing.html.

There are three conflicts there. For all of them, I think what I did is better (but maybe you think "make" is more appropriate than "makes" in the first conflict).

The last time I tried to resolve a conflict (clicking "resolve conflicts"), things went downhill quickly, so I'd prefer you resolve them. When clicking "resolve conflicts", the first is the new one (`<<<<<<< intent .... ======`)

davidcarlisle:
sorry that's what i meant:
(just one file with diff)

PR #417, opened 2022-08-19
Proofreading Chapter 3 (part 1)

brucemiller:
A beginning to proofreading Chapter 3.

I'm downplay entities, at least rephrasing so the "should"s don't imply you have to use entities rather than characters. I'm tempted to suggest removing all mention of entities, other than a mention of the Entity spec somewhere in Chapter 2.  If we do show entities, shouldn't they be displayed as `&ApplyFunction;`? Is that possible with CSS before/after on class `entity`?

I also patched up the section structure which seems to have gotten nested wrong. Hopefully I've got it right.

And a few minor edits for brevity along the way.

### DIFF
--- a/src/fundamentals.html
+++ b/src/fundamentals.html
@@ -720,7 +720,7 @@
 
    <p>Authors wishing to encode white space characters at the start or end of
    the content of a token, or in sequences other than a single space, without
-   having them ignored, must use <code class="entity">nbsp</code> (U+00A0)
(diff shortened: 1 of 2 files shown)


davidcarlisle (approved):
> shouldn't they be displayed as &ApplyFunction;? Is that possible with CSS before/after on class entity?

Yes, probably better to do a global edit across all the files and add `&amp;` ad `;` in the source rather than rely on css (as we do, mostly, for class=element where `&lt;..&gt;` is in the source.)

could be done after this round, so ths looks good to merge

Issue #233, opened 2021-01-03
<interval> has ambiguous interpretations as both an expression and qualifier


Jamesernator:
The interval element as specified can be used as both [an expression](https://mathml-refresh.github.io/mathml/#parsing_interval.class) or as [a qualifier](https://mathml-refresh.github.io/mathml/#parsing_DomainQ). This leads to ambiguity when combined with other operators.

For example when used with union (and intersection, etc), it's unclear whether to match `<interval>` as part of `DomainQ` or not without knowing the intent of it's usage:

```html
<apply>
  <union/>
  <interval>
    <cn>2</cn>
    <cn>3</cn>
  </interval>
  <interval>
    <ci>x</ci>
    <ci>y</ci>
  </interval>
</apply>
```

I think that this ambiguity is fairly confusing and that it would probably be a good idea to deprecate `<interval>` usage as a qualifier and instead to use `<domainofapplication><interval>` for such cases (or `<lowlimit>` and `<uplimit>`) e.g.:

```html
<apply>
  <int/>
  <bvar><ci>x</ci></bvar>
  <!-- definitely a qualifier, no ambiguity -->
  <domainofapplication>
    <interval>
      <cn>2</cn>
      <cn>3</cn>
    </interval>
  </domainofapplication>
  <ci>
</apply>
```

davidcarlisle:
We don't have plans to change Content MathML elements this time (MathML 4.0 focusing on integrating MathML-Core and  the `intent` attribution mechanism).

That said, I agree this could be clearer. I believe the intention is that `<interval>` is always interpreted as a qualifier  if it immediately follows a `<bvar>`  and it is always interpreted  as an interval constructor in all other cases. So there is technically never an ambiguity.

However the spec has never said that explicitly just hints at it with

https://w3c.github.io/mathml/spec.html#contm_domainofapplication_qualifier

> Alternatively, the interval qualifier could be used with an explicit bound variable:

So I don't think we can remove or rename `<interval>`-as-qualifier, but we could  say as above it is only a qualifier if following `<bvar>`, and that using  domainofapplication or lowlimit/uplimit might be less confusing.  @samdooley ? @brucemiller ?

PR #418, opened 2022-08-20
clarify interval use for #233

davidcarlisle:
Add additional text and links to clarify when `<interval>` is interpretd as a qualifier.  For issue #233 . I do not believe this changes the intended intrpretation from MathML 1,2,3.


brucemiller (approved):
This isn't in my comfort zone, but I think this clarification is correct & needed.
So, count this as approval unless @samdooley objects.

PR #420, opened 2022-08-23
Proofreading pass of section 5.1 Intent

brucemiller:
Reorganizing, rewording, condensing, clarifying.

You may need to look at an html version to appreciate.


davidcarlisle (approved):
I fixed up some html `<h4>...</h5>`  and changed the comment about starting names back to `_` from `-` (hyphen isn't an NCName start character) otherwise looks good, merging.

PR #424, opened 2022-09-16
update status of MathML in the world

NSoiffer:


bkardell:
I guess this is because we are about to ship a third/final implementation? If so, I wonder if there is value in noting that this support is mathml-core subset, specifically and authors should check on the current status?

NSoiffer:
@bkardell: it was a kind of pointless paragraph saying that MathML in html/xhtml means rendering in browsers, and then went on to mention some earlier (Opera) CSS rendering or using JS. It was kind of off-topic and definitely out of date independent of the chromium implementation.

davidcarlisle (approved):

PR #427, opened 2022-09-28
Underscore examples for language-near intent values

### LINKED ISSUE #422: Intent should prefer reference over repeating a literal
In the intent example "x prime" the intent is given as
`intent="x-prime"`.

I understand that this will work, but I'd like to include a strong suggestion that
a better way is `intent="$a-prime"` , where `arg="a"` has been added to the `<mi>x</mi>`.
It is generally bad practice to repeat the literal instead of referencing it.  If the `x`
were changed to another letter, then the intent would also have to be changed.
(Use of literal makes sense in cases like the Bell number occurring a couple
of examples later.)

That example also contains an error:  the default reading writes `f superscript...` instead of `x superscript...`.

https://github.com/w3c/mathml/blob/8971e48e83c489723841394b78bba84cc58800e7/src/mixing.html#L293

dginev:
Fixes #422 .

Adding some examples for using the underscore function to build up language-near descriptions in the intent values, as per [my comment in #422](https://github.com/w3c/mathml/issues/422#issuecomment-1244171833). 

I stopped short of recommending a certain style is _better_, since there is usually a qualifier of "better if you are in context X and want to achieve benefit Y". That could probably one day be a note of some sort.

But I introduced some visual effects that I am already unhappy about - my local rendering shows a whole screen of centered x primes, which can be a bit disorienting.

I ended up making the branch in the main repository, so no live gh-pages preview - apologies. You'll have to do a checkout and run `jekyll serve` locally to see the changes.

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -298,18 +298,19 @@
 <blockquote>x prime<br/> x superscript prime end superscript</blockquote>
 </div>
 
-<p>Language-near intent values can also be annotated compositionally, via the underscore function.</p>
+<p>Custom accessible descriptions, such as author-preferred variable or operator names, can also be annotated compositionally, via the underscore function.</p>
+<p>The above notation could instead intend the custom name "x-new", which we can mark with a single literal `intent="x-new"`, or as a compound narration of two arguments:</p>
 <div class="example mathml mmlcore">
 <pre>
 &lt;msup intent="_($base,$script)">
   &lt;mi arg="base">x&lt;/mi>
-  &lt;mo arg="script" intent="_prime">&#x2032;&lt;/mo>
+  &lt;mo arg="script" intent="_new">&#x2032;&lt;/mo>
 &lt;/msup>
 </pre>
-<blockquote>x prime</blockquote>
+<blockquote>x new<br>x superscript prime end superscript</blockquote>
 </div>
 
-<p>Chunking may also be a clear style for cases where fragments are explicitly localized. A cyrillic (Bulgarian) example:</p>
+<p>Using the underscore function may also add clarity when the fragments of a compound name are explicitly localized. A cyrillic (Bulgarian) example:</p>
 <div class="example mathml mmlcore">
 <pre>
 &lt;msup intent="_($base,$script)">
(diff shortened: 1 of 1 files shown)


dginev:
Actually, going to my fork and rebasing it on this branch after the fact is a very easy thing to do.

So you can now find a live preview of the PR here:
https://dginev.github.io/mathml/

davidcarlisle:
>  run jekyll serve 

Note people don't need to install jekyll, any web server will do,  I tend to use

`python3 -m http.server 8080`

but a local IIS or apache etc would also work. (but yes thanks the fork is easier for people without a local server)

src/mixing.html
```diff
@@ -298,7 +298,39 @@ <h4 id="mixing_intent_examples">Intent Examples</h4>
 <blockquote>x prime<br/> x superscript prime end superscript</blockquote>
 </div>
 
-    <p>Similarly an over bar may represent complex conjugation, or mean (average), again with possible readings with and without <code class="attribute">intent</code>:</p>
+<p>Language-near intent values can also be annotated compositionally, via the underscore function.</p>
```
davidcarlisle: I'm not sure "Language-near intent values can also be annotated compositionally" is understandable without having been in multiple zoom calls previously. Space isn't at a premium, Perhaps spell it out more;

The intent `_` function, which does not generate speech, may be used to compose intent expressions which are close to  natural language.

src/mixing.html
```diff
@@ -298,7 +298,39 @@ <h4 id="mixing_intent_examples">Intent Examples</h4>
...
+<pre>
+&lt;msup intent="_($base,$script)">
+  &lt;mi arg="base">x&lt;/mi>
+  &lt;mo arg="script" intent="_prime">&#x2032;&lt;/mo>
+&lt;/msup>
+</pre>
+<blockquote>x prime</blockquote>
+</div>
+
+<p>Chunking may also be a clear style for cases where fragments are explicitly localized. A cyrillic (Bulgarian) example:</p>
```
davidcarlisle: Actually I don't know what you mean by chunk here, can we spell this out more?

davidcarlisle (approved):
I commented on the wording, but wording in the whole chapter needs review anyway so we could merge first. However I will leave for a bit so it is easier for others to see the diff and perhaps comment.  I'll merge after this evening's call if not before.

dginev:
@davidcarlisle I tried to rephrase a little, also refactored with a "custom name" example to reveal more of the possible uses. 

But I may need more trial-and-error to get the right level of detail and clarity. Thanks for iterating with me.

davidcarlisle:
@dginev unless you have more imminent changes I suggest we merge now and any more wording changes can follow from general future edits

dginev:
Sounds good, thanks David.

PR #429, opened 2022-12-17
Update fundamentals.html

Zearin:
- Mark up table header cells as such
- Fix erroneous `class` attribute value for `code` element (near the beginning)
- A couple of minor punctuation copy-edits

### DIFF
--- a/src/fundamentals.html
+++ b/src/fundamentals.html
@@ -77,11 +77,11 @@
     </pre>
    </div>
 
-   <p>HTML does not support namespace extensibility in the same way, the HTML parser
-   has in-built knowledge of the HTML, SVG and MathML namespaces. <code class="attribute">xmlns</code> attributes are
-   just treated as normal attributes. Thus when using the HTML serialisation of MathML,
+   <p>HTML does not support namespace extensibility in the same way. The HTML parser
+   has in-built knowledge of the HTML, SVG, and MathML namespaces. <code class="attribute">xmlns</code> attributes are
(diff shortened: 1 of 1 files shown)


src/fundamentals.html
```diff
@@ -409,7 +409,7 @@ <h5 id="fund_units">Length Valued Attributes</h5>
 
     <p>In addition, the attributes on <a href="#presm_mpadded">mpadded</a>
      allow three <dfn>pseudo-units</dfn>, <code>height</code>,
-      <code>depth</code> and <code>width</code> (taking the place of one of the usual CSS units)
+      <code>depth</code>, and <code>width</code> (taking the place of one of the usual CSS units)
```
dginev, line 412: Do we have an official position on the use of the [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma) in the specification?

Should we?

davidcarlisle: @dginev I don't think we do, or are necessarily consistent. It does appear elsewhere in the spec, eg

>  Identifiers can include variables, function names, and symbolic constants

I don't like it much, despite living in Oxford, but I can live with it either way.

davidcarlisle (approved):

davidcarlisle:
@dginev  I think this got lost in internal discussions. OK to merge?

dginev:
> @dginev I think this got lost in internal discussions. OK to merge?

Certainly OK to merge. I was mostly fishing for clarification - which I got - thanks!

davidcarlisle:
@Zearin   thanks

PR #438, opened 2023-02-23
Intent properties

davidcarlisle:
This is a draft of chapter 5 with additions covering some of the changes discussed in parallel intent Issues.

Specifically it adds `:` for isa/properties and has wording around white space, empty head self references and error recovery.

It's made a draft PR so it doesn't get merged, but does provide a useful UI to see/comment on the diffs, and also the rendered html respec view which is available  at

https://davidcarlisle.github.io/mathml/#mixing_intent_grammar

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -69,7 +69,7 @@
    <p>The value of the <code class="attribute">intent</code> attribute, should match the following grammar.</p>
 
    <pre class="def bnf">
-intent             := term property* hint? | property+ hint? | application 
+intent             := S ( term property* hint? | property+ hint? | application ) S 
 term               := concept-or-literal | number | reference 
 concept-or-literal := NCName
 number             := '-'? digit+ ( '.' digit+ )?
@@ -244,6 +244,23 @@ S                  := [ \t\n\r]*
      read as an inter-word space.</p>
   </section>
 
+  <section>
+   <h4 id="mixing_intent_self">Intent Self References</h4>
+   <p>The grammar allows the intent to omit the leading <i>term</i>
+   and just consist of a non-empty list of properties. This should be
+   interpreted as specfying properties for the current element. This
+   can be a useful technique, especially for large constructs such as tables as
+   it allows the children to be inferred without needing to be
+   explicitly referenced in the `intent` as would be the case with an
+   <i>applicaton</i>.
+   For example, `&lt;mtable intent=":array">...` might read the table as
+   an array of values, and `&lt;mtable intent=":aligned-equations">...`
+   might read the table in a style more appropriate for a list of
+   equations. In both cases the navigation of the underlying table
+   structure can be supplied by the AT system, as it would for an
+   un-annotated table.</p>
+  </section>
+  
   <section>
    <h4 id="mixing_intent_errors">Intent Error Handling</h4>
    <p>An intent processor may report errors in intent expressions in
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -266,8 +266,8 @@ S                  := [ \t\n\r]*
    <p>An intent processor may report errors in intent expressions in
    any appropriate way, including returning a message as the
    generated text, or throwing an exception (error) in whatever form
-   the implementation supports. However in  web platform context it is
-   often not appropriate to report errors to the readers who have no
+   the implementation supports. However in  web platform contexts it is
+   often not appropriate to report errors to the reader who has no
    access to correct the source, so intent procesors should offer a
    mode which recovers from errors as described below.</p>
    <section>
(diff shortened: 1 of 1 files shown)


brucemiller:
Overall, I like it. My only quibble would be the separation of `hint` and `isa`.  Syntactically, there is no distinction between them, and semantically only minor distinction since `hint`s come from a short specified list.  It adds a extra noise character to the syntax. It could be annoying in forcing an order (properties before hints?), or alternatively a more complicated grammar.

OTOH, it handily restricts to a single `hint`.  But not really, as @dginev showed, you can get around this restriction by wrapping extra annotated `mrow`s.  So, we probably should specify which one wins, in either case: first|outer vs last|inner.

I'll unstrategically admit that this isn't a deal-breaker for me; I just think it would be cleaner & more flexible to combine them.

davidcarlisle:
@brucemiller  I'm tempted by combining `:` and `@` but there are some benefits to having them separate, hints are a fixed list and we can say (more or less) what each does, and the grammar can make sure only one is active.  isa/properties are somewhat more variable you can chain them and (I think) at the moment they can only be documented as "might have an effect, one day". It's not impossible these descriptions could be combined but might take some care. If @NSoiffer adds (or has already?) some support for`:` in mathCAT we could get a better feel for how this works in practice. 

I think it's easier to combine later (just change `@` to `:`) than separate later (detect which `:` need to be `@`) so I went for separation at this  time

NSoiffer:
MathCAT accepts `:` but doesn't do anything with it yet. The current MathCAT grammar requires `:` before `@`. Obviously, I'll update the grammar MathCAT accepts once we get closer to agreement.

Maybe someone has a philosophy as to how multiple "types" are suppose to be used and I missed it. It seems a bit foreign to me since the closest programming language analogy I can think of is inheritance, in which case one would write something like
```
<mi intent=":time:si-unit:unit">ms</mi>
```
Is order relevant?

Having multiple units complicates using them (both in parsing them [a string is no longer sufficient, and checking for a desired value), so I hope there is a strong case for them.

brucemiller:
An example like yours is exactly what I was thinking of, in which case order wouldn't matter.

davidcarlisle:
This is still miles from final form, but it seems the group will want to add `:property` syntax in some form, so I think overall it will be less confusing to merge this now so everyone is looking at the same version. So I removed the `draft` status.

dginev:
I may have leaned towards making the spacing/empty allowances even more restricted.
<details>
Here is a limit case example of an undesirable expression I could think of that the PR makes grammatical:

```xml
intent="                                                  factorial





(                                                                             8
:natural-number
    ,_:_:_:_(_:_)(_)(:_)()(),
 :_,
               :_@infix                             ,
_)
        "
```

This is the sort of syntax I was thinking of w.r.t whitespace questions being a possible fit for "error recovery".

With spaces dropped, the line noise I wrote tightens down to:
```xml
intent="factorial(8:natural-number,_:_:_:_(_:_)(_)(:_)()(),:_,:_@infix,_)"
```

which also contains a range of other unproductive abuses of the grammar allowances. I am not suggesting we can avoid all of it, just that some (new) grammatical cases come to mind that may be error-like in nature.

P.S. A very well-defended AT system may even succeed to pronounce this intent string as "eight factorial", but I don't think I reasonably expect that.

</details>

davidcarlisle:
@dginev  yes although there is not much we can do if someone does `_(_,__,___,_,)` The grammar is using "empty head" for self references, and allowing that also in arguments. Depending what we end up doing about that we could make parts of your example invalid and not allow empty heads at all, or only allow them at the outermost level, but currently allowing some flexibility for experiments. That would anyway only stop a small part of your example.

davidcarlisle:
@NSoiffer  should I merge this?
https://davidcarlisle.github.io/mathml/#mixing_intent_grammar

NSoiffer:
```
application        := intent S '(' S arguments? S ')'
arguments          := intent ( S ',' S intent )*
```
This doesn't seem right. I thought people were in favor of dropping zero argument `applications`. I haven't caught up on all the comments though. Maybe this is a "allow this because we are trying to minimize illegal syntax" arguments?

Since `intent` has `S` at its start and end, I don't think you need them around `arguments?`. Nor do they need to be in the RHS of `arguments`.

brucemiller:
I left out empty funcall arguments in #446 by accident, and you seemed to ask for them back.  I was never actually against them, however. (although they only really make sense in programming contexts?)

NSoiffer:
Ok. Then other than the seemly extra spaces, I'm ok with merging this so we have something more up-to-date to look at.

davidcarlisle:
> Since intent has S at its start and end, I don't think you need them around arguments?. Nor do they need to be in the RHS of arguments.

Oh a product of too many rearrangements,  I'll tidy up before merging, thanks.

Issue #448, opened 2023-03-16
What's in a name (also "reference", etc)?


NSoiffer:
Regardless of whether we go with the template syntax or the function syntax, there is an ongoing discussion about what characters are allowed in a name. This issue is here to focus on that topic and pulls in a thread from #446 which starts at [this comment](https://github.com/w3c/mathml/issues/446#issuecomment-1471392091).

Note: the MathML full meeting today agreed that we should pull out `number` as a specific terminal that uses `.` as the decimal separator, so that is not in question . The discussion is focused on `concept-or-literal`, `reference`, `property`, `literal`, and `name` (depending on the version of the spec).

To maybe summarize some comments:
* being freer with what is allowed means less errors that can be made
* for at least literals, speech engines won't know what to do with non letters/digits (e.g., U+2345 "⍅") so results could be unexpected
* reference names should be like an `xml:id` (without the restriction of them being unique)
* Unicode defines a regexp for ["identifier names"](https://unicode.org/reports/tr31/)
* any of the characters allowed (`#`, `@`, etc) are characters that can't be used for some future extension of the grammar

brucemiller:
With regard to your second point, are
```
<mo>⍅</mo>
<mo intent="⍅">x</mo>
```
expected to be treated differently by AT? (Offhand, I'd expect the same treatment)

brucemiller:
While I prefer permissive to restrictive, I agree that it reserving some characters for future use is a good idea.  Candidates might include the categories: keyboard-top-row, punctuation, fences, quotes.  To the extent those categories make sense and to the extent that they wouldn't reasonably be needed in normal mathematical phrases.

dginev:
I'd split the issue into two sub-components:
1. `property` and `concept` are the two caregories that need name-specific considerations (`:system-of-equations`, `binomial-coefficient`). I have always been fond of dash-separated names for that as in
   ```ebnf
   IntentName  := Letter (Letter | '-' Letter)*
   ```

3. The other categories Neil enumerated may be closer to system data than natural names (such as `xml:id` or `number` formats). Bruce [pointed out](https://github.com/w3c/mathml/issues/446#issuecomment-1471992031) that reference/arg values fit closer to ids than names.

davidfarmer:
> <mo>⍅</mo>
> <mo intent="⍅">x</mo>
> 
> expected to be treated differently by AT? (Offhand, I'd expect the same treatment)

On the call I attempted to get clarification on that point.
[...]

PR #452, opened 2023-03-19
symbolic intents

davidcarlisle:
as discussed in #448 this PR adds a new `symbol` class for intent along with a warning that perhaps it shouldn't be used until software catches up.

### DIFF
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -118,6 +118,11 @@ S                  := [ \t\n\r]*
      A literal <a data-link-for="intent" data-link-type="dfn" href="#intent_number">number</a> such as <code>2.5</code> denotes itself.
     </dd>
 
+    <dt><dfn data-dfn-for="intent" id="intent_symbol">symbol</dfn></dt>
+    <dd>
+     A literal <a data-link-for="intent" data-link-type="dfn" href="#intent_symbol">symbol</a> such as <code>+</code> denotes itself. See <a href="#mixing_intent_symbol warning"></a>.
+    </dd>
+
     <dt><dfn id="intent_reference">reference</dfn></dt>
     <dd>
     An argument [=reference=] such as <code>$name</code> refers to a descendent element
--- a/src/mixing.html
+++ b/src/mixing.html
@@ -70,9 +70,10 @@
 
    <pre class="def bnf">
 intent             := S ( term property* | property+ | application ) S 
-term               := concept-or-literal | number | reference 
+term               := concept-or-literal | number | symbol | reference 
 concept-or-literal := NCName
 number             := '-'? \d+ ( '.' \d+ )?
+symbol             := [\pSm]+
 reference          := '$' NCName
 application        := intent '(' arguments? S ')'
 arguments          := intent ( ',' intent )*
@@ -81,7 +82,9 @@ S                  := [ \t\n\r]*
    </pre>
 
    <p>Here <a href="https://www.w3.org/TR/REC-xml-names/#NT-NCName"><code>NCName</code></a>
-   is as defined in  in [[xml-names]], and <code>digit</code> is a character in the range 0–9.</p>
+   is as defined in  in [[xml-names]],
+   <code>\pNd</code> is a character of class Nd (decimal digit),
+   <code>\pSm</code> is a character of class Sm (math symbol).</p>
 
 
    <p>The parts consist of:</p>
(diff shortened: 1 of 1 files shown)


davidcarlisle:
rendered view at https://davidcarlisle.github.io/mathml/#mixing_intent_grammar

NSoiffer:
Two things:
1. I think it would be better to omit the space before the property as then it is clearer that the property belongs to what precedes as opposed to sitting on its own. I understand space is allowed -- this is for clarity sake for those first seeing the spec.
2. The discussion for why literal strings are bad does not get at what to me is the problem with them: they don't take in the needs of the reader. I think this can be made clearer by adding to `semi-factorial:postfix($x)` saying it might be read as "x semi factorial" by if "x" were a simple variable, but a blind user needs to know where something starts and ends. A more complex expression such as `x+2`, then it might be read as "quantity x plus 2 end semi-factorial". And then reference this when talking about problems with literals (e.g., if `_semi-factorial` had been used). This may not be a good example though because someone would probably use parens for "x+2", so some bracketed quantity such as "norm" is better.  But that's not postfix.  (Also, `$n` is maybe better for the `semi-factorial` example)

davidcarlisle:
@NSoiffer 

> I think it would be better to omit the space before the property as then it is clearer that the property belongs to what precedes as opposed to sitting on its own. I understand space is allowed -- this is for clarity sake for those first seeing the spec.

Do you mean omit the `S` in the grammar,  or mean in examples do `foo:infix` not  `foo :infix` ?

> The discussion for why literal strings are bad does not get at what to me is the problem with them:

Yes I agree but that text isn't touched by this PR (or the one just merged that combined `@`and `:`) so it could be separately editorially improved on `main` or a separate PR.

NSoiffer:
The later: `foo:infix` in the examples. At first glance, the `:infix` is not attached to anything. `foo:infix` makes it much clearer that "infix" is attached to "foo".

davidcarlisle:
ah OK then, I have no objection to changing the examples in the spec, but there are no  examples using `:`  in the text added by this PR,

brucemiller:
I'm not sure I see the advantage of adding this `symbol` clause. Only adding math symbols makes sense on the surface, but only slightly reduces the Error footprint. The caveat implies that the AT may or may not actually look at the symbol but may (or may not?) pass it directly on to some underlying speech-engine layer.  I wonder if anyone who hasn't participated in these discussions will understand that distinction.

I suspect I'd prefer sticking to more restrictive literals (if I had to choose).

src/mixing.html
```diff
@@ -81,7 +82,9 @@ <h4 id="mixing_intent_grammar">The Grammar for <code class="attribue">intent</co
    </pre>
 
    <p>Here <a href="https://www.w3.org/TR/REC-xml-names/#NT-NCName"><code>NCName</code></a>
-   is as defined in  in [[xml-names]], and <code>digit</code> is a character in the range 0–9.</p>
+   is as defined in  in [[xml-names]],
+   <code>\pNd</code> is a character of class Nd (decimal digit),
```
dginev, line 86: was `\pNd` meant to replace `\d` in the `number` rule? It hasn't yet.

davidcarlisle:
@brucemiller 
> I suspect I'd prefer sticking to more restrictive literals (if I had to choose).

yes I might too, not sure yet, but I made the PR to have something concrete to shoot down.

I suspect my main use of this  is to enable (or almost enable) one possible explanation of what `$op` means as in  #449 

I'd like to say it somehow causes the implied intent of the element  `op`  to be included.

If the element does not have an `intent`  then the implied  intent could/should be its content which  means `<mo>†</mo>` has an implied intent of  `†` which does not mean that we _have_ to allow this in the literal syntax `<mo intent="†">†</mo>` but it might help the explanations if it were possible to expand all `$argref` and have a valid `intent`.

Or we could instead say the implied intent of `<mo>†</mo>` is its default reading, not its content, so `<mo intent="dagger">†</mo>`  but that makes the expansion of `$argref` not just interpretation of properties and choice of speech, system dependent.

src/mixing.html
davidcarlisle: @dginev er yes, or the other way round. I oscillated on what form to use, and seem to have used one and described the other...

davidcarlisle: fixed

davidcarlisle:
This PR served its purpose of providing an explicit proposal and allowing discussion of symbolic intents, but I don't feel any consensus that it's  a good idea, so I'm going to close this.

PR #462, opened 2023-04-07
table and script properties

davidcarlisle:
This PR implements a first draft of the changes agreed on the call of 2023-04-06

 * empty head property lists are restricted to the top level so `intent=":property (:property)"`is invalid

 * The description of `:infix` fixity properties pulled out of the grammar section to a new "core properties" section which also describes table  properties and properties on sub/sup scripts.


NSoiffer:
Looks good to me. I will do the merge.

Issue #463, opened 2023-04-07
Split  Chapter 5?


davidcarlisle:
`intent` started (a _long_ time ago) as `semantics`, a more manageable inline annotation as an alternative to using parallel xref linked presentation and content mathml trees.

As such the original "mixing" chapter 5 got split with 5.1 on `intent=` and 5.2 on `<semantics>`

As the focus has moved more towards accessibility I'm not sure chapter 5 makes sense as  a chapter.

We could split it, having chapter 5 on semantics and a new chapter 6 on intent (or  the other way round, or even make intent chapter 4, after the presentation mathml chapter).

Apart from meaning we no longer have to have a "unifying" introductory section, it means we can lift all the intent sections up a level.  We are already using `<h5>` and it would be good to avoid going deeper.

davidcarlisle:
meeting agreed to this split: will be closed by #464

PR #464, opened 2023-04-11
Intent chapter

davidcarlisle:
First steps to splitting chapter 5 as suggested in #463 

a rendered view is available at https://davidcarlisle.github.io/mathml/#mixing_intent


NSoiffer:
Although I very much like the idea of lifting the sections up a level, I'm not as keen on creating a new chapter for this. If we do, I think more thought needs to be given as to how to make this more about accessibility, potentially bringing in the accessibility appendix. Apologies about not responding to #463 before you did this work.

On the other hand, if we are going to provide an initial "core" lists of intent names and properties, then it will definitely be large and warrant being on its own.

Let's discuss this at the meeting on Thursday if we have time after the AT discussions.

davidcarlisle:
> before you did this work.

Yes that's why I didn't actually change any words, just globally changed h4 to h3 and h5 to h4, Just to get something concrete to look at.

So no actual work done here, no problem to delete this branch if we decide not to do #463

PR #466, opened 2023-05-03
Distinguish literal from unknown concept

brucemiller:
This PR is a modest alternative to #459 that primarily distinguishes concepts not *currently* know by the AT and literals; the latter now being only explicit literals prefixed by `_`.

Even though an unknown concept is *treated* the same as an explicit literal, there is an important distinction: an unknown concept is only unknown for a particular AT and a particular version of the concept dictionaries; future updates may add it to a dictionary and then it is known, possibly with special behaviors.  
So, while I would generally discourage use of literals, where someone insists on specific speech that doesn't correspond to a sensible concept, I would recommend they use a literal. This seems better than using a concept  hoping that it probably never will be added to a dictionary.  Of course, if *they* think it's a concept then a concept it is.

In this PR, I'm  making the distinction in the text below the grammar rather than in the grammar itself.  This is solely to avoid getting mired in character class details.  If we decide to drop `NCNAME` and be more explicit in the grammar (something along the lines of a concept starting with a letter; a literal starts with an "_"), we probably should have separate `Concept` and `Literal` productions.


src/intent.html
```diff
@@ -376,7 +384,13 @@ <h3 id="mixing_intent_warning">A Warning about [=literal=] and [=property=]</h3>
      the meaning behind the expressions become more opaque,
      and thus excessive use of these features will tend to limit the AT's ability
      to adapt to the needs of the user, as well as limit translation and locale-specific speech.
-     Thus, the last two examples would be discouraged.
+     Thus, the last two examples would be discouraged.</p>
+     <p>Conversely, when specific speech not corresponding to a meaningfuil concept
```
davidcarlisle: meaningfuil

src/intent.html
```diff
@@ -87,23 +87,32 @@ <h3 id="mixing_intent_grammar">The Grammar for <code class="attribue">intent</co
...
+      an empty literal "_" generates no glue words;
+      effectively it implies the <code>:silent</code>
+      <a data-link-for="intent" data-link-type="dfn" href="#mixing_intent_core_properties"> property</a>.</p>
      </li>
     </ul>
     </dd>
     <dt><dfn data-dfn-for="intent" id="intent_number">number</dfn></dt>
     <dd>
-     A literal <a data-link-for="intent" data-link-type="dfn" href="#intent_number">number</a> such as <code>2.5</code> denotes itself.
+     An explicit <a data-link-for="intent" data-link-type="dfn" href="#intent_number">number</a> such as <code>2.5</code> denotes itself.
```
davidcarlisle, line 110: ah good catch, unrelated use of "literal" we should make this change in any case, it is independent of any concept/literal rewording

src/intent.html
```diff
@@ -87,23 +87,32 @@ <h3 id="mixing_intent_grammar">The Grammar for <code class="attribue">intent</co
...
+      These will never be included in an [=Intent Concept Dictionary=].
+      The reading of a literal is generated by replacing any
       <code>-</code>, <code>_</code>, <code>.</code> in the name by
       spaces and then reading the resulting phrase.
-      Names starting with <q>`_`</q> (U+00F5)  are always considered to be [=literal=] names,
-      and should never be in an [=Intent Concept Dictionary=].</p>
+      Note that while AT will often add "glue" words when generating speech
+      (such as the "of" when <code>intent="f(x)"<code> generates "f of x"),
+      an empty literal "_" generates no glue words;
+      effectively it implies the <code>:silent</code>
```
davidcarlisle: I don't like the formulation in these last 4 lines, I prefer the existing formulation (which is also preserved by this PR)

https://github.com/w3c/mathml/pull/466/files#diff-070c687a5474e331eb76ad076825db8aa124df40e0e9a292ad9157d5265720efR273


> If none of these properties is used, the
     <code>function</code> property should be assumed unless the literal is
     silent (for example <code>_</code>) in which case the <code>silent</code> property
     should be assumed. 

the current wording applies to any silent head eg `__` not just `_` and more importantly it avoids adding a new mechanism  for silent that potentially conflicts with properties.  The current wording explictly does not apply to `_:function`  as it  is just a default property, but what does "effectively" mean here?

I would re-word this so it is a forward reference to the line in the above link

something like

```
Note that while AT will often add "glue" words when generating speech
      (such as the "of" when <code>intent="f(x)"<code> generates "f of x"),
      an empty literal "_"  generates no glue words as it uses the <code>:silent</code> property,  see
    <a href="#intent_fixity_hint"></a>.
   ```

davidcarlisle (changes requested):

src/intent.html
brucemiller: thanks

src/intent.html
brucemiller: Actually, upon rereading, I think it's better to leave this out to keep the initial itemization concise.

OTOH, I think that the default fixity discussion should probably move "up" (either before our after the concept/literal `<ul>`), since it is not specific to literals but applies to concepts as well.  Perhaps it's worth making explicit that `:silent` should also omit any glue words.

davidcarlisle: OK as that has gone in last commit, OK to approve this PR.

(re-)ordering of text likely to come up anyway now intent split to its own chapter and some possible merges of text from accessbility appendix.

davidcarlisle (approved):

dginev (approved):
I am very fond of the minimal rewording here. 

This PR achieves the main remaining clarification which had ongoing discussion in #459 , and is a great substitute to the work there.

NSoiffer:
I am merging this so we have a hopefully better document to discuss at the meeting.

PR #471, opened 2023-08-24
update core concept description

davidcarlisle:
as discussed on the call today

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -199,19 +199,21 @@ S                  := [ \t\n\r]*
        name to be spoken.  For such cases, one should use an Intent
        [=property=] instead of an Intent [=concept=].  The Math
        Working Group maintain a similar list of <a
-       href="https://w3c.github.io/mathml-docs/intent-core-concepts/">list
+       href="https://w3c.github.io/mathml-docs/intent-core-properties/">list
        of core property names</a> as described in the following
        section.</p>
 
        <p>AT reading MathML attributed with a concept (or property)
        in these lists SHOULD consider this concept (or property) to be
-       a hint how the content could be read.  AT MAY use this concept
+       a hint how the content could be read.  AT systems MAY use this concept
        as a hint to improve braille generation.  However, because
        common notations have many specialized ways of being spoken AT
        is NOT constrained to use the name given.  For example, AT may
-       vocalize <q>three quarters</q>, as <q>three over x</q> (for
-       <code>&lt;mfrac></code>), or <q>three divided by x plus y</q>
-       (for <code>&lt;mo>/&lt;/mo></code>), depending on the contents
+       vocalize a fraction marked up with <code>&lt;mfrac></code>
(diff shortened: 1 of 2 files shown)


src/intent.html
```diff
@@ -176,27 +176,55 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
...
+       Group based on experience with different AT implementations and
+       following the guidelines set out in [[?Concept-Lists]].
+       
+
+       <p>Some readings benefit from annotating the kind of
+       mathematical object, rather than giving an explicit concept
+       name to be spoken.  For such cases, one should use an Intent
+       [=property=] instead of an Intent [=concept=].  The Math
+       Working Group maintain a similar list of <a
+       href="https://w3c.github.io/mathml-docs/intent-core-concepts/">list
```
dginev: Is this intentionally the same link? Could we at least use a `#properties` placeholder for the part of the page that will hold the properties? So one of:

```
https://w3c.github.io/mathml-docs/intent-core-concepts/#properties
```
or, if meant to be a separate note
```
https://w3c.github.io/mathml-docs/intent-core-properties
```

src/intent.html
```diff
@@ -176,27 +176,55 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
...
+       of core property names</a> as described in the following
+       section.</p>
+
+       <p>AT reading MathML attributed with a concept (or property)
+       in these lists SHOULD consider this concept (or property) to be
+       a hint how the content could be read.  AT MAY use this concept
```
dginev: `AT MAY` would be more legible as `AT systems MAY` here. 

It is easy to misread when the two separate all-caps terms touch.

src/intent.html
```diff
@@ -176,27 +176,55 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
...
+       as a hint to improve braille generation.  However, because
+       common notations have many specialized ways of being spoken AT
+       is NOT constrained to use the name given.  For example, AT may
+       vocalize <q>three quarters</q>, as <q>three over x</q> (for
```
dginev: I believe the word "as" before `three over x` shouldn't be there.

dginev (approved):
Looks great in general - I left a few comments that may need minor edits.

src/intent.html
davidcarlisle: sorry I'll fix, it was supposed to be https://w3c.github.io/mathml-docs/intent-core-properties

src/intent.html
davidcarlisle: done

src/intent.html
davidcarlisle: I think it's a bit over abbreviated and hard to understand I'll adjust and ping you once commited,

dginev:
Looks even better, thanks again.

PR #474, opened 2023-10-07
Intent Chapter Property lists and tables

davidcarlisle:
As discussed on the call on 2023-10-05 the two commits in this PR adjust the properties section to reference the lists in mathml-docs, and adjusts the table examples removing the editors note saying we plan to do something and adding the properties as already shown at https://w3c.github.io/mathml-docs/intent-examples/#id-2-3f2dea5b3d4622d90c3db6fa446e7f0c

The diff is shown here or the rendered sections at

Properties:  https://davidcarlisle.github.io/mathml/spec.html#mixing_intent_properties

Table examples: https://davidcarlisle.github.io/mathml/spec.html#mixing_intent_examples_mtr


NSoiffer:
I'll do the merge, but I do have some suggestions for further fixes:
1. "such as on the head of an [application](https://davidcarlisle.github.io/mathml/spec.html#intent_application)" -- I think it is clearer to say "... head of an intent [application]....". I had to stop and think, "what application is this talking about?" Of course the link resolves that question, but I shouldn't need to click to the link to figure out the sentence.
2. I think the "prime" examples are poor examples. The default reading should be "x prime" not "x superscript prime end superscript", and that makes the examples not that interesting. Maybe a superscript of "-1" (inverse vs power vs superscript/end superscript)? Or maybe a bar over x?

dginev:
I would prefer if we avoided the recommendation:

> As existing AT already implements relatively complex
       mechanisms to navigate tabular alignments, it is often not
       desirable to try to force explicit speech forms via <code
       class="attribute">intent</code> attributes making use of
       concepts or literal terms, however by use of properties the
       author can give hints to the speech generation and generate
       speech suitable for a list of aligned equations rather than say a matrix.
       Some examples are given below.

Instead, it would be better if we recommended:

> Whenever a kind of tabular construct has an associated Core Property, 
it is usually better to use *only* the property and allow AT to infer 
how to navigate the expression.
>
> In the Open realm of annotation, tabulars which specify a layout of math fragments
 can have high complexity for navigation and speech generation. 
> Such cases can benefit from dedicated intent expressions, 
> backed by careful testing with concrete AT systems.

For any non-standard tabular (especially in the research document realm) there is no existing AT support (beyond simply walking the table cells), and intent annotations can certainly be an improvement. I had a couple of examples at the top of #337 to illustrate such tabulars (a diagram, and a grammar definition).

NSoiffer:
I mostly agree with Deyan's reformulation for the first paragraph, but "navigate" is not the right thing to say (it wasn't right in the first formulation, but I missed that). Perhaps "...allow AT to infer how to _speak and_ navigate the expression".

I'm not keen on the second paragraph though. Maybe something more like
> When core properties are insufficient to represent a tabular layout, the use of intent concept names and, if appropriate, also properties from the open list of properties should be used to convey the desired speech and navigation of the tabular layout. Because of the likely complexity of these layouts, testing with AT should be done to verify that users hear the expression as the author intended.

davidcarlisle:
I'll add something along the lines of the above (as this is more editorial re-wording than adding new section, I'll probably just push to main, but I'll ping here for review

davidcarlisle:
@dginev  @NSoiffer  A merge of all three versions and a back link to `#intent_table_properties` where these properties areintroduced.

Issue #209, opened 2020-05-06
Use fence/separator operator properties in MathML Core?


fred-wang:
cc @Joanmarie @asurkov 

They are currently mentioned here:
https://mathml-refresh.github.io/mathml-core/#operator-fence-separator-or-accent-mo
https://mathml-refresh.github.io/mathml-core/#dictionary-based-attributes
https://mathml-refresh.github.io/mathml-core/#operator-dictionary

However, these are not used at all for layout purpose.

They should be keep in MathML Core if they are necessary for assistive technologies (and so browsers must expose them via their accessibility tree). I believe Gecko and WebKit only expose them for NSAccessibility but I'm not even sure that's used by VoiceOver. In Chromium this is an open question at https://bugs.chromium.org/p/chromium/issues/detail?id=1042665

They are also discussions elsewhere about how to expose more math semantics, so maybe any method decided there could apply for fence/separator too instead of MathML-specific attributes. But that seems out of the scope of MathML Core.

fred-wang:
Multichar entries of the operator dictionary are now handled by https://mathml-refresh.github.io/mathml-core/#operator-dictionary-compact

For the record, current estimated size is 770*2 = 1540 bytes. The cost of supporting fence/separator is significant enough, (65+3)*2 = 136 bytes so 9% of the dictionary size.

Probably not a big deal, rendering engines won't implement this anyway. This would only be needed by browsers if the AT really need them and if that can't be done by other means than dictionary-based attributes.

fred-wang:
I think there were mistakes in my previous script. Anyway, most recent estimate can be found in the spec.

NSoiffer:
For what it is worth, MathPlayer does not make use of either of those properties to speak or navigate math.

On the other hand, chrome is running at about 1,500,000,000 bytes right now on my computer; worrying about 136 bytes seems like a complete waste of time.

fred-wang:
> On the other hand, chrome is running at about 1,500,000,000 bytes right now on my computer; worrying about 136 bytes seems like a complete waste of time.

This is the important part from my previous comment: "rendering engines won't implement this anyway. This would only be needed by browsers if the AT really need them and if that can't be done by other means than dictionary-based attributes.".  So I expect this is something to be answered by a11y team / AT developers, not something decided by the CG. If they don't need it, we should just remove this from MathML Core.

fred-wang:
[...]

PR #500, opened 2024-08-21
hide fence and separator

davidcarlisle:
as agreed in the call of 15-08-2024 this hides the `fence` and `separator`  properties for issue #209


NSoiffer:
You might have gone a little overboard. I see you removed `<code class="attribute">accent</code>,`. in 
```
when they exist, are supported using the Boolean attributes <code class="attribute">fence</code>,
    <code class="attribute">separator</code> and <code class="attribute">accent</code>, which can be used to distinguish these cases.</p>
```
I know that's not in core, but it is still in full so maybe this was too aggressive a removal?

davidcarlisle:
@NSoiffer  yes I wondered about that, the paragraph didn't have any normative statement and it didn't seem to read well once I removed two of the three attributes it mentioned so I dropped the whole thing. If we want to say something about `accent` at that point I think it would need a completely new paragraph specifically about that rather than the existing general text about unstated "subtle properties" which only made sense when discussing the longer list, I thought.

NSoiffer:
Given that it's not in core, maybe the less said about the default values, the better.

Issue #501, opened 2024-08-23
actiontype attribute required on maction?


davidcarlisle:
In previous versions `actiontype` is a required attribute on `<maction>`

The MathML4 draft spec still says it is required. https://w3c.github.io/mathml/#presm_maction_att

That table lists `actiontype` and `selection` as "not-core"

The MathML4 draft schema makes both attributes optional
https://w3c.github.io/mathml/#parsing_maction

as it inherits from the mathml4-core schema,

Despite the full spec flagging the attributes as not-core, the core spec does describe both attributes (but with no defined effect).

So either the full spec should make the attribute optional or the mathml-presentation schema needs to over-ride the imported mathml-core schema and make this required in full.

Given that the behaviour of maction is largely now undefined, I think making the full spec match the schema and match core by making `actiontype`  optional is probably best.

This shows up in test/test-core-03.xml in the mathml-schema repository.

davidcarlisle:
At the call on 29-08-2024 resolved to consider deprecating maction completely at a future meeting

NSoiffer:
The 5-9-2024 call discussed deprecation. Everyone agreed that one can't specify an action in a platform independent way, so in some sense the element is pointless. Others felt that `maction` provides a place to hang actions on, whether they are in JS or done some other way -- it serves as a way for authors to indicate something should be done there. It also provides a place for data to be added that some code can act on (e.g., showing the steps in deriving an equation).

A straw poll was conducted and 2 people wanted it deprecated, three want it kept (aligned with core), and one person was neutral.

davidcarlisle:
closed by #502

PR #502, opened 2024-08-24
make actiontype optional to match core #501

davidcarlisle:
This adjusts the description of `actiontype` so that it is not mandatory on `maction` it also uses data- attributes rather than a local namespace in the example. As noted in #501 mathml-core and the current mathml4 full schema draft make actiontype optional already.

The new rendering is available at

https://davidcarlisle.github.io/mathml/#presm_maction


davidcarlisle:
@NSoiffer in the call just now we discussed going further and deprecating maction but I'd like to merge this PR anyway to bring the spec into line with the schema and make actiontype optional attribute.

davidcarlisle:
The WG call on 2024-09-12 resolved to merge this

Issue #428, opened 2022-11-03
Deprecate <none> element


dginev:
I have recently started to wonder if the `<none>` element has outlived its usefulness.
- The current MathML Core [description of `<none>`](https://w3c.github.io/mathml-core/#dfn-none) states:
   > The empty `<mprescripts>` and `<none>` elements are laid out as an `<mrow>` element.
- The naming of `<none>` appears inconsistent with the rest of Presentation MathML, as one would have expected `<mnone>`.
- MathML Core also states that "[Unknown MathML elements](https://w3c.github.io/mathml-core/#dfn-unknown-mathml-element) behave the same as the [`<mrow>`](https://w3c.github.io/mathml-core/#dfn-mrow) element"
    - which induces backwards compatibility in a world where `<none>` is relegated to unknown, and `<mrow>` is recommended in its place.

---

**Proposal:** Deprecate and ultimately remove `<none>`, in favor of using an empty `<mrow>` element in its place.

---

P.S. There may be reasons to preserve `<none>` that I am not yet aware of - comments are most welcome. I also suspect this suggestion comes too late for MathML 4, and will only be addressed in work following that milestone.

NSoiffer:
> The current MathML Core [description of <none>](https://w3c.github.io/mathml-core/#dfn-none) states:
> The empty <mprescripts> and <none> elements are laid out as an <mrow> element.

That description only applies when they are not in an `mmultiscripts`. Prior to that statement, it says:
>  ...then they are laid out according to the CSS specification where the corresponding value is described. Otherwise, the layout below is performed. 

I don't know whether your proposal presents a problem with that description.

An empty `mrow` is sometimes used to indicate a slot to be filled in, so potentially it means something different than `none`.

`none` is also used by elementary math in `msrow`, `mscarries`, and `mscarry` to indicate an empty column.  These uses might be compatible with your proposal, but I'd need to think about it more.

As for why it is "none" and not "mnone", I don't remember. There are a few other elements that don't start with 'm' that are allowed in presentation: `semantics`, `annotation`, and `annotation-xml`. One could argue they straddle the line between presentation and content MathML and hence aren't strictly part of presentation MathML.

As for timing, I think now is the time to make the change if people think it is appropriate: do it before core is an official rec. This probably should be a MathML core issue.

dginev:
> That description only applies when they are not in an mmultiscripts. Prior to that statement, it says:
[...]

PR #506, opened 2024-09-15
deprecate none for #428

davidcarlisle:
This replaces `<none/>` by `<mrow/>`  as discussed on #428

### DIFF
--- a/src/changes.html
+++ b/src/changes.html
@@ -88,15 +88,18 @@
     implementation infer the row markup if it was omitted.
     </li>
     <li>The use of <code class="element">malignmark</code> has been
-      restricted and simplified, matching the features implemented in
-      existing implementations.
-      The <code class="attribute">groupalign</code> attribute on table
-      elements is no longer supported.</li>
+    restricted and simplified, matching the features implemented in
+    existing implementations.
+    The <code class="attribute">groupalign</code> attribute on table
+    elements is no longer supported.</li>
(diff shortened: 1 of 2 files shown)


NSoiffer:
For `mmultiscripts`, you wrote that `mrow` or any other empty element can be used as a replacement for `none`. That's not entirely true (e.g, `<mglpyh/>` and  `<malignmark>` aren't legal), so maybe a small rewording is in order.

You correctly do _not_ say that for the elementary math case, but since that section is widely separated from `mmultiscripts`, I think it the phrasing you used earlier bears repeating in that section:
> (The element <code class="defn emptytag">&lt;none/&gt;</code> was used in earlier MathML releases, but was equivalent to an empty <code class="emptytag">&lt;mrow/&gt;</code>).

davidcarlisle:
@NSoiffer  wording adjusted.

For mmultiscripts I say that any (sensible) empty element couldl be allowed (in practice it would have to be mrow-like such as mstyle, perhaps mtext

For mscarries (as the text calls out the empty case to _not_ wrap the child in an implies mscarry) I say it has to be an empty mrow to get that handling.

I hope that's OK An alternative would be to say for mmultiscripts that mrow has to be used, but that's not really enforcable, any element allowed as a child of mmultiscripts might potentially be empty

NSoiffer:
Nice fix (adding "valid").

Issue #304, opened 2022-03-31
Potential presentation MathML items to deprecate in MathML 4


NSoiffer:
#303 is about previously deprecated items. This is about presentation MathML items that we should think about deprecating in MathML 4 based on not been used. The "not being used" part is based on my experience and needs to be verified.

Here is a list of potential items to deprecated based on usage:

- [x] mspace@linebreak -- was added for compatibility with MathML when MathML 3's linebreaking was added. It's a wart and I don't think it is used (not that MathML 3's linebreaking was used outside of DSI's tools and the polyfil I wrote).
- [x] malignmark/malgingroup -- these were added to allow sub-alignment in a table. They are very hard to implement and few systems have done so.
- [x] maction -- this really should be done with JavaScript. At the very least, we should pull some of the example `actiontype` values such as `statusline`, `tooltip`, and `input`.
- [x] menclose -- remove `radical` notation (#3); maybe also remove longdiv since it too is duplicative

davidcarlisle:
One other (that I think I would remove raher than deprecate is 

- [x] namespaced atributes for undefined behaviour.

Not specifying these will have no effect on anything except validators, and validators can choose to use mathml4-legacy schema that makes them valid (currently they are valid in the draft full schema, but  I think we could change that).   Currently (inherited from core) mathml-presentation allows `data-*` attributes  I think we should allow that in Content as well then drop `other` and namespaced attributes.

dginev:
This comment confirms the keywords mentioned  [in David's comment](https://github.com/w3c/mathml/issues/294#issuecomment-1079885285) from issue #294 are indeed not used by mainstream conversion tools in early 2022.

The attributes themselves are quite rare, and what little use I spotted was with numeric values (largely in `em` or `px` units, if anyone is curious).

---

From the full conversion engines I surveyed: latexml, tex4ht, tralics, pandoc, plastex. They do not seem to emit the attribute-keyword pairs in question.

From the formula conversion tools, TeXzilla and katex have no reliance on the keywords.

From what I can see, the [MathJax-src](https://github.com/mathjax/MathJax-src) repository has most thorough coverage on the subject, but the code is already defensive - e.g. in "MathML 2 compatibility" input guards, or in test files. I don't see reliance on the keywords for present-day output.

I used local grep and github's code search functionality to conduct this survey - hopefully no major oversights.

NSoiffer:
At the June 19 meeting we agreed to:
1. Deprecate linebreak on mspace
[...]

Issue #181, opened 2020-01-02
MathML 4 extensions for alignment and possible deprecation of  <maligngroup/> and <malignmark/>


NSoiffer:
Having just implemented a polyfill for elementary math, that got me thinking about some related ideas:
1. The most obvious concept related to long division is synthetic division. It is basically the same idea as long division except that you are dividing polynomials. With synthetic division, the columns contain numbers (the coefficients of the polynomial), not just digits.  As a refresher, see [this page](https://www.purplemath.com/modules/synthdiv.htm) and the example taken from it below):

Polynomial Division | Synthetic Division
---------------------- | -------------------
![image](https://user-images.githubusercontent.com/1545836/71687830-811e3f00-2d53-11ea-8dc6-06ce2cc598fe.png) |  ![image](https://user-images.githubusercontent.com/1545836/71687868-8b403d80-2d53-11ea-941a-6f1ba1d8d3c0.png)

2. Synthetic division is a shorthand for long division of polynomials (left example above). Long division of polynomials is basically the same idea as long division of numbers except that instead of digits, you have monomials that need to go into their own column. Doing that automatically requires knowing the variable you want to "sort" on so that each monomial goes into the proper column.
3. A very similar property is needed when displaying systems of equations -- each monomial wants to be in it's own column (in this case, the top level element would not be `mlongdiv`, but `mstack`). 

There are a few complications such as decimal alignment of the coefficients:
```
      8.44x + 55  y =  0
      3.1 x -  0.7y = -1.1
```
Note that alignment requires knowing what characters/operators act as column separators (e.g., `+`and `-`, along with `=` and a few other relational operators).  These would be inside of `mo` elements, so potentially any `mo` element could be a separator, or maybe an attribute specifies what the separators are (something to think about/discuss).

The above example is taken from the [MathML 3 spec for`maligngroup` and `malignmark`](https://www.w3.org/TR/MathML3/chapter3.html#presm.malign). I think only MathPlayer ever implemented those elements and I suspect that you can count on your fingers the number of times they have been used. It is a very complicated feature to implement and to use. In contrast, I think the above features are an incremental extension to elementary math layout, so implementation (especially via an extension to the polyfill I wrote), means that supporting these features would be universal (assuming I or someone else extended the polyfill). Just as important, using this extension would be simple as it is a declarative notation that doesn't require modifying the generated layout other than at a high level (wrapping with `mstack` or `mlongdiv`). It would be less powerful though.

[...]

Issue #25, opened 2019-02-22
Remove/Deprecate mglyph element


fred-wang:
Currently this is not defined in MathML Core, only MathML 4.
- I think the need for custom glyph is already covered with modern HTML5 technologies by using web fonts and "Private Use Areas".
- If people want to use image, they can embed an HTML img tag into an mtext.

In any case, I think the idea of mglyph is a rather hacky temporary workaround. Ideally, people use characters defined in Unicode. These days, they should have enough characters available to write math.

physikerwelt:
discussed in the telco on 2019-02-25:
* patrick ion points out that generating new glyphs is something which is used heavily in mathematics, .e.g, in knot theory.
* embedding HTML would be a more powerful option and superseded the mglyph functionality

pdfion:
There was apparently a technology proposed for CJK gaiji that would do just fine for supplemental math characters: SING and replace <mathglyph>. I have no idea if this is viewed as a current technology.  Anyone know?
References:
* https://en.wikipedia.org/wiki/OpenType#SING_gaiji_solution
** The SING (Smart INdependent Glyphlets)[81][82] technology that made its debut with Adobe's Creative Suite 2 allows for the creation of glyphs, each packaged as a standalone font, after a fashion. Such a packaged glyph is called a glyphlet. The format, which Adobe has made public, is based on OpenType.   
* SING: Adobe’s New Gaiji Architecture,Jim DeLaHunt (2004) | http://partners.adobe.com/public/developer/en/opentype/SING_Intro_IUC26.pdf

pdfion:
That should read SING could replace `mglyph`.

pdfion:
`mglyph`

fred-wang:
@pdfion Thanks for the information. I think someone should investigate more, maybe contact people from CSS / Font WG to see if that's the kind of features already available in browsers ( https://drafts.csswg.org/css-fonts-3/ ). In any case, that aligns with the idea that this is something that has to be provided by CSS/font so we should just rely on it instead of defining our own feature in MathML.

(PS: I've fixed the formatting in your comment)

dani31415:
While it is somehow acceptable that MathML4 is MathML3 + CSS - deprecation of attributes, if we remove the `mglyph`, we loss the capability of inserting images in MathML (for the case that the MathML is not part of an HTML 5).

I suggested Stand-alone MathML profile #58. Just in case we forgot that MathML should also exist outside web pages.

[...]

PR #507, opened 2024-09-18
"not core" notes for mglyph and malignmark

davidcarlisle:
See #181 #304 #25

rendered version at

https://davidcarlisle.github.io/mathml/#notcore-mglyph


davidcarlisle:
This is non controversial, I'll merge to simplify diffs for mlabeledtr

PR #510, opened 2024-10-12
remove other

davidcarlisle:
Other is removed from the common attributes in presenttaion and content
It is replaced by namespaced attributes (as in mathml3) and the mathml4 dtat-* attributes
The conformance section updated
Change appendix updated
Schema updated (at https://github.com/w3c/mathml-schema) but would display in appendix A to remove other
from the schema except mathml4-legacy which restores it. 
https://github.com/w3c/mathml-schema/commit/20bba04b80445c3cde30e919b94882e41bffff23

### DIFF
--- a/src/changes.html
+++ b/src/changes.html
@@ -40,6 +40,11 @@
     replaced by suitable use of <code class="attribute">display</code>.
     The `mathml4-legacy` schema makes these valid if needed for legacy applications.</li>
 
+    <li>Remove the <code class="attribute">other</code> attribute.
+    This have been deprecated
+    since MathML&#160;2.
+    The `mathml4-legacy` schema makes this valid if needed for legacy applications.</li>
+
    </ul>
   </section>
 
--- a/src/conformance.html
+++ b/src/conformance.html
@@ -162,7 +162,7 @@
    <section>
     <h5 id="interf_deprec">Deprecated MathML 1.x and MathML 2.x Features</h5>
 
-    <p>MathML 3.0 contains a number of features of earlier MathML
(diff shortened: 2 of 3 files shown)


NSoiffer:
I don't see why you kept the reference to "other". 
```
   <p>The attribute <code class="attribute">other</code>
   was <a class="intref" href="#interf_deprec">deprecated</a>
   in MathML&#160;2 and has been removed from MathML&#160;4.
```
Haven't we just deleted all the other features that have been removed for over a version?

NSoiffer:
Also, good catch on all the MathML 3 references you found....

davidcarlisle:
> I don't see why you kept the reference to "other".
> 
> ```
>    <p>The attribute <code class="attribute">other</code>
>    was <a class="intref" href="#interf_deprec">deprecated</a>
>    in MathML&#160;2 and has been removed from MathML&#160;4.
> ```
> 
> Haven't we just deleted all the other features that have been removed for over a version?


I have a similar statement about mlabelledtr forst para pf

https://w3c.github.io/mathml/#presm_eqno

We could review all these but I don't think it harms to say it's been removed in the appropriate section, as well as in the change appendix.

NSoiffer:
`mlabeledtr` was part of MathML 3, so it makes sense to mention that it has been removed. `other` hasn't been part of MathML for two decades. There are many other features that have been removed that aren't mentioned. One that pops into mind is `mglyph`s `index` and `fontfamily` attrs which were part of MathML 2.

davidcarlisle:
OK I'll zap

davidcarlisle:
@NSoiffer  "other" paragraph gone

PR #511, opened 2024-10-22
Define what a concept match is (plus some cleanup)

NSoiffer:
The main change is to make another attempt at clarifying when a core or open (if also part of the AT's dictionary) is a match. This PR is an alternative to https://github.com/w3c/mathml/pull/477.

I did _not_ generalize "match" to include the looser concept of matches caused by training some NMT (neural machine translation). The concepts are the same, although obviously NMTs do not do dictionary lookups -- the training data would essentially be the result of a dictionary lookup.

I also made a number of editorial changes including fixing some spelling errors, and unifying wording.

There is, I think, still a problem that we use "Intent Concept Dictionary" and also use the word "list". This might make sense because we provide core and open _lists_, but the implementation looks things up in a _dictionary_ to see if there is a match. I tried to be consistent in the text about this, but I don't believe the text makes this clear.

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -90,7 +90,7 @@ S                  := [ \t\n\r]*
       A <dfn id="intent_known_concept">known concept</dfn> matches a name
       in an [=Intent Concept Dictionary=] recognized by the AT.
       This may produce specific audio or braille renderings based on the speech hints
-      given in the dictionary.
+      given in the list.
       An <dfn id="intent_unknown_concept">unknown concept</dfn> is a concept not
       currently known to the AT. These will be treated the same as a [=literal=], spoken as-is.
       However, future updates of the AT and [=Intent Concept Dictionary=] may
@@ -162,19 +162,20 @@ S                  := [ \t\n\r]*
   <section>
     <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
      <p>An <dfn>Intent Concept Dictionary</dfn> is a mapping from a [=concept=] name
-      to specific speech or braille for that concept.
+      to speech or braille for that concept.
       The mapping may take into account any [=property=] that follows the name.
       AT that makes use of <code class="attribute">intent</code>
       SHOULD be able to produce speech or braille that corresponds to any
-      of the concepts in the [=Core=] table discussed below.
+      of the concepts in the [=Core=] list discussed below.
       AT that makes use of <code class="attribute">intent</code>
-       MAY also include concepts in the [=Open=] table discussed below,
+       MAY also include concepts in the [=Open=] list discussed below,
        as well as its own built-in dictionaries.
       </p>
       <p>The Intent Concept Dictionary is somewhat analogous to the <a href="#oper-dict"></a> used by
       MathML renderers in that it provides a set of defaults renderers should be aware of.
       The <code>property</code> also has some analogies to the operator dictionary's use of
-      <code class="attribute">form</code>.
+      <code class="attribute">form</code> because a match makes use of arity properties
+      (<code>prefix</code>, <code>infix</code>, etc.).
     </p>
     <!--<p class="issue" data-number="410">Issue 410</p>-->
     <p>
@@ -207,12 +208,12 @@ S                  := [ \t\n\r]*
        in these lists SHOULD consider this concept (or property) to be
        a hint how the content could be read.  AT systems MAY use this concept
        as a hint to improve braille generation.  However, because
-       common notations have many specialized ways of being spoken AT
+       common notations have many specialized ways of being spoken, AT
        is NOT constrained to use the name given.  For example, AT may
        vocalize a fraction marked up with <code>&lt;mfrac></code>
        as <q>three quarters</q> or <q>three over x</q>
        or may vocalize an inline fraction marked up as <code>&lt;mo>/&lt;/mo></code>
-       as <q>three divided by x plus y</q>.
+       as <q>three divided by x</q>.
        The choice may depend on the contents
        and carrier element associated with an
        <code>intent="divide($num,$denom)"</code>.</p>
@@ -223,11 +224,12 @@ S                  := [ \t\n\r]*
       who can not see the a fraction, AT might say <q>fraction x over
       three end fraction</q> so the listener knows exactly what is
       part of the fraction.  For someone who can see the content,
-      these extra words can be a distraction.  AT should always
+      these extra words might be a distraction.  AT should always
       produce speech that is appropriate to the community they serve.
       </p>
       </li>
-      <li><dfn id="intent-open">Open</dfn>: This is an <a href="https://w3c.github.io/mathml-docs/intent-open-concepts">open list of concepts</a>
+      <li><dfn id="intent-open">Open</dfn>: This is an
+      <a href="https://w3c.github.io/mathml-docs/intent-open-concepts">open list of concepts</a>
       to which contributions are invited.
       AT reading MathML attributed with a name in this list
       MAY use the speech hints provided by the intent
@@ -245,9 +247,19 @@ S                  := [ \t\n\r]*
      <a data-cite="INFRA#ascii-case-insensitive">ASCII case-insensitive</a>
      and also normalize
      <q>`_`</q> (U+00F5) and <q>`.`</q>  (U+002E) to <q>`-`</q> (U+002D).
-     If the speech hints are not being used
-     and the concept name is being read then each of  `-`, `_` and `.` should be
-     read as an inter-word space.</p>
+     An intent matches an entry in the core or open concept list when the normalized name, the fixity property
+     (which may be defaulted in the concept dictionary), and the number of arguments all match.
+     If a match is found, the speech hint in the list
+     should be used be used as a guide for the generation of suitable text
+     for the <a href="#intent_known_concept">known concept</a>.
+     If the intent does not match any entries, then it is treated as
+     an <a href="#intent_unknown_concept">unknown concept</a>.
+     Unknown concept names should be read as a literal after normalizing each of `-`, `_` and `.`
+     to an inter-word space.
+    Both <a href="#intent_known_concept">known concepts</a> and
+    <a href="#intent_unknown_concept">unknown concepts</a> should be read in a manner
+    consistent with any given or default arity properties.
+  </p>
   </section>
 
   <section>
@@ -294,7 +306,7 @@ S                  := [ \t\n\r]*
       <li>
       In the case of a [=known concept=] name, the property MAY be used in choosing the alternatives
       supported by the AT. For example <code>union</code> is in the
-      Core dictionary with speech patterns "$1 union $2" and "union of $1 and $2".
+      Core list with speech patterns "$1 union $2" and "union of $1 and $2".
       An intent <code>union :prefix ($a,$b)</code> would
       indicate that the latter style is preferred.
       </li>
@@ -328,7 +340,7 @@ S                  := [ \t\n\r]*
      should be read in style suitable for matricies, with typically
       column numbers being announced.</li>
       <li>`system-of-equations` should be read in style suitable for
-      displayed equations (and inequations), with typically
+      displayed equations (and inequalities), with typically
       column numbers not being announced. Each table row would
       normally be announced as an "equation" but a
       `continued-equation` property on an <code>mtr</code> indicates
--- a/src/intent.html
+++ b/src/intent.html
@@ -253,7 +253,6 @@ S                  := [ \t\n\r]*
      should be used be used as a guide for the generation of suitable text
      for the <a href="#intent_known_concept">known concept</a>.
      If the intent does not match any entries, then it is treated as
-     an <a href="#intent_unknown_concept">unknown concept</a>.
      Unknown concept names should be read as a literal after normalizing each of `-`, `_` and `.`
      to an inter-word space.
     Both <a href="#intent_known_concept">known concepts</a> and
--- a/src/intent.html
+++ b/src/intent.html
@@ -174,7 +174,7 @@ S                  := [ \t\n\r]*
       <p>The Intent Concept Dictionary is somewhat analogous to the <a href="#oper-dict"></a> used by
       MathML renderers in that it provides a set of defaults renderers should be aware of.
       The <code>property</code> also has some analogies to the operator dictionary's use of
-      <code class="attribute">form</code> because a match makes use of arity properties
+      <code class="attribute">form</code> because a match makes use of fixity properties
       (<code>prefix</code>, <code>infix</code>, etc.).
     </p>
     <!--<p class="issue" data-number="410">Issue 410</p>-->
@@ -258,7 +258,7 @@ S                  := [ \t\n\r]*
      to an inter-word space.
     Both <a href="#intent_known_concept">known concepts</a> and
     <a href="#intent_unknown_concept">unknown concepts</a> should be read in a manner
-    consistent with any given or default arity properties.
+    consistent with any given or default fixity properties.
   </p>
   </section>
 
--- a/src/intent.html
+++ b/src/intent.html
@@ -247,7 +247,7 @@ S                  := [ \t\n\r]*
      <a data-cite="INFRA#ascii-case-insensitive">ASCII case-insensitive</a>
      and also normalize
      <q>`_`</q> (U+00F5) and <q>`.`</q>  (U+002E) to <q>`-`</q> (U+002D).
-     An intent matches an entry in the core or open concept list when the normalized name, the fixity property
+     An intent matches an entry in a concept list when the normalized name, the fixity property
      (which may be defaulted in the concept dictionary), and the number of arguments all match.
      If a match is found, the speech hint in the list
      should be used be used as a guide for the generation of suitable text


davidcarlisle (approved):

src/intent.html
```diff
@@ -245,9 +247,19 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
...
-     If the speech hints are not being used
-     and the concept name is being read then each of  `-`, `_` and `.` should be
-     read as an inter-word space.</p>
+     An intent matches an entry in the core or open concept list when the normalized name, the fixity property
+     (which may be defaulted in the concept dictionary), and the number of arguments all match.
+     If a match is found, the speech hint in the list
+     should be used be used as a guide for the generation of suitable text
+     for the <a href="#intent_known_concept">known concept</a>.
+     If the intent does not match any entries, then it is treated as
+     an <a href="#intent_unknown_concept">unknown concept</a>.
```
dginev: >  If the intent does not match any entries, then it is treated as an <a href="#intent_unknown_concept">unknown concept</a>.

I would rephrase this to:

> When an intent does not match any entries in AT-supported lists, the expected baseline  behavior is that for an <a href="#intent_unknown_concept">unknown concept</a>.

src/intent.html
```diff
@@ -245,9 +247,19 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
...
+     Unknown concept names should be read as a literal after normalizing each of `-`, `_` and `.`
+     to an inter-word space.
+    Both <a href="#intent_known_concept">known concepts</a> and
+    <a href="#intent_unknown_concept">unknown concepts</a> should be read in a manner
+    consistent with any given or default arity properties.
```
dginev: Are "arity properties" defined somewhere? Unsure what they are.

dginev (commented):
Just a couple of minor edits.

src/intent.html
davidcarlisle: <strike>I don't think they are defined,. How about .... `consistent with the number of arguments provided in this instance of the intent` ?</strike>

src/intent.html
davidcarlisle: I think Neil's wording is better. A concept not matching the lists known to the system is literally the definition of `unknown-concept` so the fact that these are treated as unknown concepts is by definition not just an expected behaviour.

```
An <dfn id="intent_unknown_concept">unknown concept</dfn> is a concept not
      currently known to the AT. These will be treated the same as a [=literal=], spoken as-is.
      However, future updates of the AT and [=Intent Concept Dictionary=] may
      include additional concepts, at which time those concepts may also receive special treatment
```

dginev: There is a subtle difference:
1. "then it is treated as" **prescribes** the 'unknown concept' behavior. 
2. "the expected baseline behavior is" signals AT can **choose to do more**.

The subtle difference is also in the text you quoted - "not currently known to the AT" is implementation-specific, and is in no way equivalent to "not in the Core or Open lists".

My two old examples where this is important were:
- a hypothetical ChemistCat AT which has its own (Open-like) list of chemistry concepts and speech hints. 
- a ML (machine learning) system which does not employ explicit lists, but is trained on a large Open domain of intent values.

I am (and have been) intentionally attempting to get a different emphasis across in the wording here.

davidcarlisle: but that difference does not exist.  If the concept does not match then it is treated as an unknown concept,. that is not a prescription of any behaviour (you can still use it as you suggest) it is simply restating the definition of unknown concept.

> The subtle difference is also in the text you quoted - "not currently known to the AT" is implementation-specific, and is in no way equivalent to "not in the Core or Open lists".

Certainly "not known to the AT" is not the same as "not in core or open lists" I do not think any text has suggested that is the case.  The open list has _no_ effect on the behaviour at all, it is a list of suggestions that implementers may like to add to their system specific lists.

davidcarlisle: > a hypothetical ChemistCat AT which has its own (Open-like) list of chemistry concepts and speech hints.

That is more or less a canoncal example of a system specific list, nothing here in any way suggests you can not do that.

    a ML (machine learning) system which does not employ explicit lists, but is trained on a large Open domain of intent values.

We had similar issues with openmath content dictionaries that were more or less documented as finite lists (of xml elements in that case) but as they could be (and were) generated on demand by some systems, they could allow a lot more than a first reading might suggest.

Your system could (depending on exactly what it does) have a concept dictionary that matches every NCName. If that's what you document that it does, then you match any name (and arity) and make whatever speech you want.

dginev: I am not sure I can restate this any more cleanly, but let me try again:

If a concept does not match _a list's speech hints_ (= the situation described by the current text), that is not the same as the concept not being _known to an AT_ (an "unknown concept"). To coin a clarifying term, we are discussing an "unmatched concept", not an "unknown" one.

The definition of "unknown concept" is not directly applicable here and **prescribing that it is** is a design choice.

And I object to that choice, because I want to protect the open experimental "gray area" between intent uses clearly prescribed in lists, and intent uses that are entirely unknown to an AT system. The space in between needs to be inhabited by experiments, so that the lists (especially the Open lists) can continue to grow when experiments stabilize.

davidcarlisle: > If a concept does not match a list's speech hints (= the situation described by the current text), that is not the same as the concept not being known to an AT (an "unknown concept"). To coin a clarifying term, we are discussing an "unmatched concept", not an "unknown" one.

Yes they are the same. An unknown concept is (by definition) a name not in the dictionary supported by the AT in use.
Such a concept is called an unknown concept and the behavior for unknown concepts is prescribed as reading them as literals.

If you have a AT system with a dynamic concept dictionary that matches every NCName then you have no unknown concepts and can do prettty much anything with any concept name.

dginev: > An unknown concept is (by definition) a name not in the dictionary supported by the AT in use.

Not the current definition, which is one I quite like: `"not currently known to the AT"`

An example I've given before (in a discussion with Moritz) is that if an AT implements a small subset of the Open list -- and only that -- any concept it hasn't yet covered would be an "unknown concept" to the AT, despite being listed. Which is _why_ we could expect the "read as-is" graceful fallback behavior in partial implementations of any list.

"unknown concept" has operational semantics _internal to an AT_, and is not a global list-related property.

I do not understand the dogmatic defense of the prescriptive text here. My change keeps the baseline behavior exactly the same, it only leaves the door open that an AT can do more when desired.

davidcarlisle: > Not the current definition, which is one I quite like: "not currently known to the AT"
yes, by that definition.

> any concept it hasn't yet covered would be an "unknown concept" to the AT, despite being listed. 

what is listed in the open list has no bearing on what is an unknown concept. It is as I said, just a list of suggestions to implementors should they wish to make anything known.

An unknown concept is a concept not in the list of names _implemented by the system in use_  Most of the names in the open list are unknown concepts in most systems. Despite the fact that they are of course listed in the open list.

> I do not understand the dogmatic defense of the prescriptive text here. 

Sorry I think you have simply misread the definition there is no prescriptive text here. I do not think that the suggested text makes sense in a specification context.

> When an intent does not match any entries in AT-supported lists, the expected baseline behavior is that for an unknown concept

the premise of this statement is exactly the definition of unknown concept so this means

> for an unknown concept the expected baseline behaviour is that of an unknown concept.

This is tautologous and even if it were not "expected baseline behaviour" is an undefined term (I could guess a meaning) but it is not a phrase suitable for a spcification.

brucemiller: Some of this discussion is a little hard to track, and seems to lead to distributed definitions.  I think that where we have `<dfn id="intent_known_concept">known concept</dfn>` (currently top of the page), as much as possible we should have a fairly clear and complete definition. It might be worth making it more explicit that "known" means the *union* of Core, Open *and* any other implicit dictionaries supported by the AT in question and is thus system specific.

davidcarlisle: > It might be worth making it more explicit that "known" means the union of Core, Open and any other implicit dictionaries supported by the AT in question and is thus system specific.

As currently defined it is _just_ those supported by the system not the core or open lists, the complement  to unknown which is defined in the current daft as:

> unknown concept is a concept not currently known to the AT

This list _should_ include all of core, but if you are using a system that does not implement all the core list then some core concepts could also be unknown.

a name being in the open list does not affect whether it is known or unknown in that sense. It is just a list of suggestions to system implementors for concepts they may want to implement.

brucemiller: Ah, yes, right; good point. But then, the section that sparked this discussion `An intent matches an entry in the core or open concept list when ...` shouldn't be mentioning the open or core lists at all, but rather should refer to the AT's own concept list (whatever it may be).

davidcarlisle: @brucemiller  er yes.  how about

>      An intent matches an entry in a concept list when the normalized name, ....

brucemiller: I think that works, yes.

src/intent.html
```diff
@@ -245,9 +247,19 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
      <a data-cite="INFRA#ascii-case-insensitive">ASCII case-insensitive</a>
      and also normalize
      <q>`_`</q> (U+00F5) and <q>`.`</q>  (U+002E) to <q>`-`</q> (U+002D).
-     If the speech hints are not being used
-     and the concept name is being read then each of  `-`, `_` and `.` should be
-     read as an inter-word space.</p>
+     An intent matches an entry in the core or open concept list when the normalized name, the fixity property
```
davidcarlisle:
```suggestion
     An intent matches an entry in a concept list when the normalized name, the fixity property
```

src/intent.html
davidcarlisle: @dginev  @NSoiffer  Ah it is defined implicitly earlier:

> `because a match makes use of arity properties     (<code>prefix</code>, <code>infix</code>, etc.).`

But elsewhere these are called fixity properties rather than arity properties which seems to be a better name. I'll make an explicit commit suggestion below.

src/intent.html
```diff
@@ -162,19 +162,20 @@ <h3 id="mixing_intent_grammar">The Grammar for <code class="attribue">intent</co
...
       AT that makes use of <code class="attribute">intent</code>
-       MAY also include concepts in the [=Open=] table discussed below,
+       MAY also include concepts in the [=Open=] list discussed below,
        as well as its own built-in dictionaries.
       </p>
       <p>The Intent Concept Dictionary is somewhat analogous to the <a href="#oper-dict"></a> used by
       MathML renderers in that it provides a set of defaults renderers should be aware of.
       The <code>property</code> also has some analogies to the operator dictionary's use of
-      <code class="attribute">form</code>.
+      <code class="attribute">form</code> because a match makes use of arity properties
```
davidcarlisle:
```suggestion
      <code class="attribute">form</code> because a match makes use of fixity properties
```

src/intent.html
davidcarlisle:
```suggestion
    consistent with any given or default fixity properties.
```

NSoiffer:
I'm good with those changes.

Anything else that needs clarification?

davidcarlisle:
> I'm good with those changes.
> 
> Anything else that needs clarification?

@NSoiffer the first one I posted as a suggested change but the second I commited (I thought I did the same thing but apparently not and I commited by mistake) but as you say you are happy with both I accepted my own suggestion for the first change so all up to date now,

dginev:
@NSoiffer I just have the one edit suggestion left.

davidcarlisle:
> @NSoiffer I just have the one edit suggestion left.

which ? Do you mean "When an intent does not match any entries in AT-supported lists, the expected baseline behavior is that for an unknown concept"

Hasn't that been discussed at length already?  The wording is not suitable for a specification: it just says the behaviour for an unknown concept is that expected for an unknown concept. apart from inserting the undefined term "baseline".

dginev:
@davidcarlisle my edit suggestion stands as written - I certainly welcome constructive rewordings.

It has been "discussed" at length already by a single group member, yes. There are even strawmen advertised in place of my wording at this point, so by internet standards this has been very fruitful.

If we can't find a common path here, let's resolve it with a group vote as per procedure.
I can present my argument in a couple of minutes, hopefully in a more welcoming setting.

davidcarlisle:
> @davidcarlisle my edit suggestion stands as written - I certainly welcome constructive rewordings.
> 
> It has been "discussed" at length already by a single group member, yes. There are even strawmen advertised in place of my wording at this point, so by internet standards this has been very fruitful.
> 
> If we can't find a common path here, let's resolve it with a group vote as per procedure. I can present my argument in a couple of minutes, hopefully in a more welcoming setting.

but you haven't suggested any alternative wording or explanation of any of the words in your suggestion we really can't use "expected" or "baseline" in a specification that is defining what is expected, and without that it's just a tortology.

NSoiffer:
@dginev: maybe I'm misreading the commits for this, but I don't see anything from you when looking at the actual commits as opposed to the message thread.

I'll add this to the agenda.

Anybody: Is there a way, short of merging the pull requests, so see what the document would read as so it is easier to discuss at the meeting? Unfortunately, I don't know github well enough to do that myself.

dginev:
> but you haven't suggested any alternative wording or explanation of any of the words in your suggestion

I offered a detailed explanation at:
https://github.com/w3c/mathml/pull/511/files#r1811039104

and 
https://github.com/w3c/mathml/pull/511/files#discussion_r1811132812

My key concern is to weaken `"then it is treated as"`, which states a direct effect.
I don't want to suggest *may*, because the distinction I am trying to establish is not that it is *optional*.

Instead, I want to establish it is a kind of foundation, or fallback, or baseline in the AT process. "In the absence of other available knowledge, it is treated as" would be the informal way of stating it.

The discussion above was not the wording choice but my key argument being misdirected (due to my alleged misreading of the current spec text), so we did not really get into word smithing.

davidcarlisle:
> @dginev: maybe I'm misreading the commits for this, but I don't see anything from you when looking at the actual commits as opposed to the message thread.
> 
> I'll add this to the agenda.
> 
> Anybody: Is there a way, short of merging the pull requests, so see what the document would read as so it is easier to discuss at the meeting? Unfortunately, I don't know github well enough to do that myself.

You can see the full text in the `match` branch but not rendered (here) as html (that's the advantage of doing a PR from a fork) but I'll sync my fork and show the match branch in gh-pages there, give me a minute...

davidcarlisle:
@NSoiffer Not sure why my commits showed here (they failed anyway, I'll check and clean up. The rendered pages are as intended

w3c  https://w3c.github.io/mathml/#mixing_intent_dictionaries

fork https://davidcarlisle.github.io/mathml/#mixing_intent_dictionaries

difference:

image

davidcarlisle:
@dginev 

> My key concern is to weaken "then it is treated as", which states a direct effect.

But the statement is that unknown concepts are treated as unknown concepts so it does not state any effect that has not already been stated. It was/is just an explanatory re-iteration it has no effect on the normative text at all.

> If the intent does not match any entries, then it is treated as an [unknown concept](https://github.com/w3c/mathml/pull/511#intent_unknown_concept).

Is just a more or less exact re-statement of the definition of `unknown concept` if you change it then you conflict with the definition. You could simply delete it but having the reminder here is helpful (or at least I can not see why it is not helpful)

dginev:
> You could simply delete it but having the reminder here is helpful (or at least I can not see why it is not helpful)

I think simply deleting it would be a major improvement actually. If it is really so minor a comment, maybe that's a quick middleground win?

davidcarlisle:
> > You could simply delete it but having the reminder here is helpful (or at least I can not see why it is not helpful)
> 
> I think simply deleting it would be a major improvement actually. If it is really so minor a comment, maybe that's a quick middleground win?

we could although it's a shame to omit a helpful reminder, but omitting it is certainly an improvement over replacing it with something that has no defined meaning and can't possibly have a meaning that is consistent with the existing definitions in the draft.

davidcarlisle:
@dginev oh would you find it easier to understand the intention here if it was explicitly a reference to the definition rather than worded as a specification that just happens to mirror the existing definition, so change

> If the intent does not match any entries, then it is treated as an [unknown concept](https://davidcarlisle.github.io/mathml/#intent_unknown_concept).

to 

> Recall that any intent that does not match any entries is an [unknown concept](#intent_unknown_concept) and treated as a [literal](#literal)


??

dginev:
@davidcarlisle almost, but if we quoted the definition as-is:
> Recall that any concept not currently known to the AT is an [unknown concept](https://github.com/w3c/mathml/pull/511#intent_unknown_concept) and treated as a [literal](https://github.com/w3c/mathml/pull/511#literal).

davidcarlisle:
> @davidcarlisle almost, but if we quoted the definition as-is:
> 
> > Recall that any concept not currently known to the AT is an [unknown concept](https://github.com/w3c/mathml/pull/511#intent_unknown_concept) and treated as a [literal](https://github.com/w3c/mathml/pull/511#literal).

whatever, that means the same thing,  hang on I'll repost that using the suggested change tool....

NSoiffer:
I'm not a fan of duplication, so I agree we should delete a sentence. However, I think the place where the duplication should be removed is right after the definition of an unknown concept:
> These will be treated the same as a [literal](https://davidcarlisle.github.io/mathml/#intent_literal), spoken as-is.

That is somewhat incomplete because it only mentions the concept (it really should say 'concept name') and not how the full intent is treated as opposed to the mention where matches are described (which includes the effect of any fixity properties).

From a flow point of view, removing the sentence also makes sense because it separates out the following sentence about updates to the AT's concept dictionary causing unknown concepts to receive "special treatment" (maybe we should say 'become known concepts'). That sentence follows on nicely from the description of known concepts to the description of unknown concepts.

> If the intent does not match any entries, then it is treated as an [unknown concept](https://davidcarlisle.github.io/mathml/#intent_unknown_concept).

If anything, I would argue (maybe as David has done several times) that "treated as" can be dropped so the sentence is "...then is an unknown concept". The following sentence says what the minimal reading requirements are for unknown concepts (normalize the head and follow any fixity properties). It makes sense to have this here because the paragraph also describes the minimal requirements for known concepts.

dginev:
@NSoiffer reducing the definitional text of "unknown concept" actually makes it more mysterious. If anything we should be expanding the text there - and getting group agreement where the boundaries are. 

And if we need more categories to refer to. The "matched" vs "unmatched" distinction has to do with speech hint availability, while the "known" vs "unknown" distinction has to do with AT implementation state. Expediently bundling them together can get rather bumpy for implementers.

I'm hopeful that David's incoming change will be a great fit with my reading, for what that's worth.

src/intent.html
```diff
@@ -245,9 +247,19 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
...
      <q>`_`</q> (U+00F5) and <q>`.`</q>  (U+002E) to <q>`-`</q> (U+002D).
-     If the speech hints are not being used
-     and the concept name is being read then each of  `-`, `_` and `.` should be
-     read as an inter-word space.</p>
+     An intent matches an entry in a concept list when the normalized name, the fixity property
+     (which may be defaulted in the concept dictionary), and the number of arguments all match.
+     If a match is found, the speech hint in the list
+     should be used be used as a guide for the generation of suitable text
+     for the <a href="#intent_known_concept">known concept</a>.
+     If the intent does not match any entries, then it is treated as
```
davidcarlisle, line 255:
```suggestion
     Recall that any concept not currently known to the AT is an  <a href="#intent_known_concept">known concept</a> and treated as a <a href="#intent_literal">literal</a>.
```

src/intent.html
```diff
@@ -245,9 +247,19 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
...
+     an <a href="#intent_unknown_concept">unknown concept</a>.
```
davidcarlisle:
```suggestion
```

dginev (approved):
With the latest pending [minor change](https://github.com/w3c/mathml/pull/511/files/b66fe44c760bbc4913499d1a0da1cbb6176e75c6#r1811595793), this PR is good to merge.

Thank you for incorporating my feedback.

src/intent.html
davidcarlisle: @NSoiffer  I suggest you click the "commit suggestion" button above and call it done:-) but I just saw your last comment which crossed with me making the above, feel free to further adjust, but it's late here and I'm mostly offline tomorrow so don't wait for a review from me.

NSoiffer:
Merging and then we can discuss this further at the meeting. I'll send out email pointing out the main parts to read.

@dginev: feel free to comment on my email with any points you want people to think about.

dginev:
@NSoiffer you merged this without the last [minor change](https://github.com/w3c/mathml/pull/511/files/b66fe44c760bbc4913499d1a0da1cbb6176e75c6#r1811595793) I mentioned above.

Was that intentional? Would you like me to carry these over for the meeting discussion in a new PR, or will you add them in yourself?

NSoiffer:
I missed that. As mentioned in #512, at this point, I want to hold off on any PRs until after the meeting so that we all read the same thing.

PR #512, opened 2024-10-23
Recall unknown concept definition

dginev:
This carries over the last minor edit we agreed to with @davidcarlisle in #511 , which was left unmerged.

Which is opportune, since I spotted a typo and moved a fragment clarifying space normalization to the "unknown concept" definition.

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -92,7 +92,8 @@ S                  := [ \t\n\r]*
       This may produce specific audio or braille renderings based on the speech hints
       given in the list.
       An <dfn id="intent_unknown_concept">unknown concept</dfn> is a concept not
-      currently known to the AT. These will be treated the same as a [=literal=], spoken as-is.
+      currently known to the AT. These will be treated the same as a [=literal=], spoken as-is
+      after normalizing each of `-`, `_` and `.` to an inter-word space.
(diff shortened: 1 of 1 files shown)


physikerwelt (approved):
lgtm

NSoiffer:
Since I requested people read this for the meeting, I don't think it is a good idea to have some people read one version and some people read a different version. We should hold off merging this PR until after the meeting.

Having had another day's distance, I see a few more tweaks to make and I suspect others will also have some suggestions.

dginev:
@NSoiffer in that case I am quite happy I linked the PR in the email thread where you made the request. Having you block progress here is unfortunate.

I welcome others to read the change here as well, so that we can move forward towards a resolution.

physikerwelt:
> and some people read a different version

I see this as an editorial change and think it's more like correcting typos. Otherwise, I think it would be even more important that people read different versions. We don't need to make it overly complicated, but to me it would be easier and more efficient if we would follow a more formal process.

dginev:
@NSoiffer now that we've decided @brucemiller is rewriting this section, can we please merge here so that he - and others proofreading - have as much nuance as possible included in the text? Thanks.

davidcarlisle (approved):

PR #513, opened 2024-10-29
Rearrange Intent section

brucemiller:
This PR is a first attempt at rearranging the intent section to collect the various parts of definitions, descriptions and behaviors into single places to make the specification easier to use. Mostly, it collects the description of known, unknown, the how properties are used into a new subsection "Using Intent Concepts and Properties".

There is a gh-pages view here: https://brucemiller.github.io/mathml/spec.html#mixing_intent, although the numbering is not correct.

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -342,7 +342,7 @@ S                  := [ \t\n\r]*
    <p>Note that future updates of the AT and [=Intent Concept Dictionary=] may
      include additional concepts, at which time those concepts may also receive special treatment.</p>
 
-   <p><em>NEEDS CLARIFICATION</em>
+   <p><em>NEEDS CLARIFICATION: </em>
      In cases where the intent is missing or does not contain an explicit concept,
      the AT may be able to infer the relevant concept. fixity and arguments
      from the structure and content of the MathML.
--- a/src/intent.html
+++ b/src/intent.html
@@ -301,7 +301,7 @@ S                  := [ \t\n\r]*
      match.  If arguments were given explicitly in the <code class="attribute">intent</code>
      then their number gives the arity, and the fixity is determined from an explicit property
      or may default from the concept dictionary. Otherwise, arity is assumed to be 0.</p>
-   <p>An concept is considered a <dfn id="intent_known_concept">known concept</dfn> (to the AT)
+   <p>A concept is considered a <dfn id="intent_known_concept">known concept</dfn> (to the AT)
      when the normalized name, the fixity property, and the arity
      all match an entry in the AT's concept dictionary.
      The speech hint in the matching entry
@@ -336,8 +336,8 @@ S                  := [ \t\n\r]*
      This can be a useful technique, especially for large constructs such as tables as
      it allows the children to be inferred without needing to be
      explicitly referenced in the `intent` as would be the case with an <i>applicaton</i>.
-     For example, `&lt;mtable intent=":array">...` might read the table as
-     an array of values, whereas `&lt;mtable intent=":aligned-equations">...`
+     For example, `&lt;mtable intent=":matrix">...` might read the table as
+     an array of values, whereas `&lt;mtable intent=":system-of-equations">...`
      might read the table in a style more appropriate for a list of
      equations. In both cases the navigation of the underlying table
      structure can be supplied by the AT system, as it would for an
--- a/src/intent.html
+++ b/src/intent.html
@@ -140,7 +140,8 @@ S                  := [ \t\n\r]*
      the core properties as described below.</p>
     </dd>
 
-    <dt><dfn id="intent_property_list">self-property-list</dfn></dt>
+<!--    <dt><dfn id="intent_property_list">self-property-list</dfn></dt>-->
+    <dt>self-property-list</dt>
     <dd>At the top level, an [=intent=] may consist of just a
     non-empty list of properties. These apply to the current element
     as described in <a href="#intent_using"></a>.</dd>
--- a/src/intent.html
+++ b/src/intent.html
@@ -86,16 +86,7 @@ S                  := [ \t\n\r]*
       <p>A <dfn id="intent_concept">concept</dfn> corresponds to some mathematical or
       application specific function or concept.
       For many concepts, the words used to speak a concept are very similar to the name
-      used when referencing a concept.
-      A <dfn id="intent_known_concept">known concept</dfn> matches a name
-      in an [=Intent Concept Dictionary=] recognized by the AT.
-      This may produce specific audio or braille renderings based on the speech hints
-      given in the list.
-      An <dfn id="intent_unknown_concept">unknown concept</dfn> is a concept not
-      currently known to the AT. These will be treated the same as a [=literal=], spoken as-is
-      after normalizing each of `-`, `_` and `.` to an inter-word space.
-      However, future updates of the AT and [=Intent Concept Dictionary=] may
-      include additional concepts, at which time those concepts may also receive special treatment.</p>
+      used when referencing a concept.</p>
     </li>
     <li>
       <p>A <dfn id="intent_literal">literal</dfn> is a name starting with <q>`_`</q> (U+00F5).
@@ -152,7 +143,7 @@ S                  := [ \t\n\r]*
     <dt><dfn id="intent_property_list">self-property-list</dfn></dt>
     <dd>At the top level, an [=intent=] may consist of just a
     non-empty list of properties. These apply to the current element
-    as described in <a href="#mixing_intent_self"></a>.</dd>
+    as described in <a href="#using_intent"></a>.</dd>
 
     <!--<dt><dfn id="intent_expression">expression</dfn></dt>-->
     <dt>expression</dt>
@@ -162,14 +153,13 @@ S                  := [ \t\n\r]*
 
   <section>
     <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
-     <p>An <dfn>Intent Concept Dictionary</dfn> is a mapping from a [=concept=] name
-      to speech or braille for that concept.
-      The mapping may take into account any [=property=] that follows the name.
-      AT that makes use of <code class="attribute">intent</code>
-      SHOULD be able to produce speech or braille that corresponds to any
-      of the concepts in the [=Core=] list discussed below.
-      AT that makes use of <code class="attribute">intent</code>
-       MAY also include concepts in the [=Open=] list discussed below,
+    <p>Every AT system that supports <code class="attribute">intent</code> contains,
+      at least implicitly, an <dfn>Intent Concept Dictionary</dfn>
+      which maps [=concept=] names to speech, text or braille for that concept
+      (The details of matching and using concept names is given in [=intent matching using=]).
+      The concept dictionary of such an AT
+      SHOULD contain the concepts in the [=Core=] list discussed below;
+      It MAY also include concepts in the [=Open=] list discussed below,
        as well as its own built-in dictionaries.
       </p>
       <p>The Intent Concept Dictionary is somewhat analogous to the <a href="#oper-dict"></a> used by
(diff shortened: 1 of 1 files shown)


davidcarlisle:
You could make respec happier by deleting the `<dfn>`  (as you have done for `element` in the following item)  or referencing it (as done for `property` on the previous item)


```
Found definition for "self-property-list", but nothing links to it. This is usually a spec bug!
```

brucemiller:
Since you mention it: I had thought to make two changes to the grammar, but didn't want to go too far mixing changes.
* fold the `self-property-list` into the 1st `intent` line and remove the `dt,dd,dfn` completely, on the hopes that it is sufficiently described in the later section. It doesn't seem like it needs to called out explicitly (and is a slightly confusing name). Or, of course, simply remove the `dfn` wrapper if we'll never reference it.
* Change `S` to `spaces` to be infinitesimally more intuitive.

src/intent.html
```diff
@@ -152,7 +143,7 @@ <h3 id="mixing_intent_grammar">The Grammar for <code class="attribue">intent</co
     <dt><dfn id="intent_property_list">self-property-list</dfn></dt>
```
davidcarlisle:
```suggestion
    <dt>self-property-list</dt>
```

src/intent.html
```diff
@@ -347,37 +287,76 @@ <h3 id="mixing_intent_properties">Intent Properties</h3>
...
+     <dfn id="intent_unknown_concept">unknown concept</dfn> (to the AT)
+     and will be treated the same as a [=literal=];
+     that is, the name is spoken as-is after normalizing each of `-`, `_` and `.` to an inter-word space.
+     Even for an unknown concept, if a fixity property and arguments were given (or inferred),
+     the speech for the arguments should be composed
+     in a manner consistent with the given fixity property, if possible.</p>
+   <p>Note that future updates of the AT and [=Intent Concept Dictionary=] may
+     include additional concepts, at which time those concepts may also receive special treatment.</p>
+
+   <p><em>NEEDS CLARIFICATION: </em>
```
davidcarlisle:
```suggestion
   <p>
```

src/intent.html
```diff
@@ -347,37 +287,76 @@ <h3 id="mixing_intent_properties">Intent Properties</h3>
...
+     In cases where the intent is missing or does not contain an explicit concept,
+     or does not contain an explicitly structured intent with concept and arguments,
+     the AT may be able to infer the relevant concept. fixity and arguments
+     from the structure and content of the MathML.
+     In such cases, it should proceed as above as if the inferred intent had been given explicitly.
+   </p>
```
davidcarlisle:
```suggestion
   For example an AT system might infer  in intent of `power` given `<msup><mi>x</mi><mn>2</mn></msup>`,
   resulting in a reading such as "x squared". An explicit intent property of `:literal`  could be used to over-ride this inferred intent, and lead to a reading such as `x superscript 2`.
  </p>
```

davidcarlisle (changes requested):
On the whole it looks good, there is one respec warning, and I added a suggestion at the point you indicated one was needed. (you can choose to add these or not to your PR using the "commit suggestion" button in the github UI.)

davidcarlisle:
@brucemiller 

> fold the self-property-list into the 1st intent line and remove the dt,dd,dfn completel

I'm not sure. We ended up with that grammar after a rather tortuous iteration and it's managed to stay stable for some time now:-)

The grammar would be simpler if we dropped `self-property-list` and instead allowed `term` to be empty so that `intent=":literal"` was just a property on an empty term rather than a specific self-property, but the consensus was that allowing empty terms (everywhere) was too much so we are where we are.

As you say it's true that we could regard that as a grammar implementation detail and not make the description follow the grammar exactly but (at least in that section) I think it is clearer if it does, as it's describing the grammar rather than describing intent in general (I think....)

src/intent.html
brucemiller: Right, could add that. But there's also the case of "short-hand" intent (if that's the right term): I was also concerned with a case like
```<mrow><mi>x</mi><mo intent="plus">foo</mo><mi>y</mi></mrow>```
which could be inferred to be
```<mrow intent="plus:infix($x,$y)">...```
Should this be addressed here? (or is it addressed elsewhere already?)

**Comment by davidcarlisle** [comment]
On  `S` or `spaces` I don't mind, I was probably unconciously swayed by

https://www.w3.org/TR/xml/#NT-S

as the xml spec used S here

But if people prefer to change that, then any name is OK

**Comment by davidcarlisle** [comment]
ah yes I think we have so far avoided trying to specify how intents on operators are lifted up to have intents with arguments on inferred children (as infering the arguments proved tricky in general) perhaps you are right that just giving some examples as system specific inferred intents would be good.

maybe a new section (before the intent error handling section) on inferring intent?

**Comment by brucemiller** [comment]
Should I go ahead & create such a section, and add the suggestions we've discussed as a starting draft?

**Comment by davidcarlisle** [comment]
> Should I go ahead & create such a section, and add the suggestions we've discussed as a starting draft?

I think so, this PR is "re-arrage intent section" not for any specific change so I think it makes sense to roll in other changes so people can see the section as a whole

**Comment by dginev** [comment]
To avoid disrupting discussion here, I have posed my remaining feedback as an add-on PR to this one, visible at [brucemiller/mathml#1](https://github.com/brucemiller/mathml/pull/1).

I am grateful for the changes in Bruce's commit 9902004c7c03e565eca16bd58674d6dcdc70d965 here, which gave me a foundation to build on.

File: src/intent.html (hunk 3 above)
Comment (NSoiffer) [reply src/intent.html]: @Bruce: perhaps you could say that the two are equivalent (without inference)? But I think a more compelling example would be better -- an example where you would expect some inference. In general, an `mrow` will read as is unless some inference overrides it E.g., $|x|$ or or maybe $P(A|B)$. In the case of $|x|$, I don't think there is a markup analogous to what you wrote. I think if you wrote `<mo intent="xxx">|</mo>`, then I'm dubious AT would do a match that comes up with absolute-value  -- MathCAT wouldn't.

One thought would be add more examples to the examples section and use them to illustrate some of the edge cases.

Comment (brucemiller) [reply src/intent.html]: I just pushed a *very* dubious section as a starting point for discussion.

**Comment by davidcarlisle** [comment]
on f3f9fec it may be worth copying (or perhaps even referencing) some of the examples in the intent-examples collecttion eg

https://w3c.github.io/mathml-docs/intent-examples/#id-1-1eff6f050a4d26222d07f68068e1014a

which shows "n choose k" being inferred from nothing (unless you use the `:literal` (nee `:structure`) property to suppress that.

**Comment by brucemiller** [comment]
It may indeed be worth copying some examples, but it would help to have a better sense of where we're trying to go with this section before getting too deep. In writing the "use of" section, the case of `<mo `intent="plus"`>` didn't quite fit: there are no explicit arguments.  Maybe the containing `mrow` just gets read literally, and "plus" is spoken at the right time, without any concern of arity or fixity?  Alternatively, if it is expected that `intent="plus:infix($x,$y)"` will somehow be inferred, we probably have to say something about how.  Either way, something needs to be said, I think, just different things!

OTOH, for the case where no intent is given,  I suspect that we should say as little as possible about inferring intent to avoid limiting implementers from finding the best magic that they can.

**Comment by davidcarlisle** [comment]
> It may indeed be worth copying some examples, but it would help to have a better sense of where we're trying to go with this section before getting too deep. In writing the "use of" section, the case of `<mo `intent="plus"`>` didn't quite fit: there are no explicit arguments. Maybe the containing `mrow` just gets read literally, and "plus" is spoken at the right time, without any concern of arity or fixity? 

Yes, I think there are two possible models there. 

 * one (which is I think what the spec says now)  `<mo intent="plus">`  is a zero-arity `plus` and that may or may not have a special rule or may just get read as literal `plus`.

* the other (which might be useful but we certainly don't want to specify, leaving it up to implementatiions, is that the system decides that `plus `is infix binary, so infers its  arguments from siblings (if there are any) and basically re-writes to `intent=plus{$a,$b)` on the parent mrow.

> Alternatively, if it is expected that `intent="plus:infix($x,$y)"` will somehow be inferred, we probably have to say something about how. Either way, something needs to be said, I think, just different things!

yes although I suspect we can't say "how" without falling into old rabbit holes around which element siblings are likely arguments and which are just punctuation

> OTOH, for the case where no intent is given, I suspect that we should say as little as possible about inferring intent to avoid limiting implementers from finding the best magic that they can.

yes I think though we should say it's allowed. One real problem I had making the intent examples was that mathcat often made a good reading of the original, no-intent version, and then I struggled to add intent in a way not to mess it up.  We need to say I think that systems may infer intents either if there are no intents at all or if the intent markup is only partial with not all subterms annotated. The system  SHOULD not do that if a `:literal` property is in scope but otherwise they have free rein to do anything?

**Comment by brucemiller** [comment]
> Yes, I think there are two possible models there.
>     * one (which is I think what the spec says now)  `<mo intent="plus"`>`is a zero-arity`plus`and that may or may not have a special rule or may just get read as literal`plus`.

Yeah, this is where my thinking was gradually headed, I think.  I didn't find it addressed explicitly anywhere, but certainly implied in examples. Probably at the beginning of the Using section we need a bit of clarification along the lines that arity & fixity are only relevant if arguments are given explicitly in the `intent` form. And the effective speech hint is going to look like the existing ones, except omitting any "of", commas, arguments and such. Perhaps some concepts will want an explicit extra entry? Some won't make sense in any event (as @NSoiffer pointed out).

If we can phrase this right, we probably don't need this new section at all.
 
>     * the other ....
The possibility of making the detailed intent completely explicit belongs with the possibility of inferring a intent from scratch: we'd want to allow it, but not get into any details.

**Comment by davidcarlisle** [comment]
> The possibility of making the detailed intent completely explicit belongs with the possibility of inferring a intent from scratch: we'd want to allow it, but not get into any details.

yes, so I suppose it's a choice of whether we assume anything not forbidden is allowed, and so not mention it or whether we want to give a note that it is allowed and possibly even a an example or two.

**Comment by brucemiller** [comment]
Following the above discussion, I just added some clarification to the beginning of the Using section about matching intents which have no arguments, which (mostly) handles the common case of `<mo intent="plus">`. It may be that we need to either add `arity=0` cases to the dictionaries, or otherwise comment on *not* saying "integral of"?

Consequently, I've removed the stub section on inference (we can add more later, if needed).

File: src/intent.html [hunk 4]
```diff
@@ -347,39 +288,70 @@ <h3 id="mixing_intent_properties">Intent Properties</h3>
...
+   <h3 id="intent_using">Using Intent Concepts and Properties</h3>
+   <p>When the <code class="attribute">intent</code> attribute corresponding to a specific node
+     contains a concept component, the AT's [=Intent Concept Dictionary=] should be consulted.
+     The concept name should be normalized
+     (<q>`_`</q> (U+00F5) and <q>`.`</q>  (U+002E) to <q>`-`</q> (U+002D)),
+     and compared using <a data-cite="INFRA#ascii-case-insensitive">ASCII case-insensitive</a>
+     match.  If arguments were given explicitly in the <code class="attribute">intent</code>
+     then their number gives the arity, and the fixity is determined from an explicit property
+     or may default from the concept dictionary. Otherwise, arity is assumed to be 0.</p>
+   <p>An concept is considered a <dfn id="intent_known_concept">known concept</dfn> (to the AT)
```
davidcarlisle:
```suggestion
   <p>A concept is considered a <dfn id="intent_known_concept">known concept</dfn> (to the AT)
```

src/intent.html
```diff
@@ -347,39 +288,70 @@ <h3 id="mixing_intent_properties">Intent Properties</h3>
...
+
+   <p>In cases where the intent contains neither an explicit nor inferrable concept
+     the AT should generally read out the MathML in a literal or structural fashion.
+     However, any given [=properties=] should be respected if possible,
+     and may be useful to indicate the kind of mathematical object,
+     rather than giving an explicit [=concept=] name to be spoken.
+     This can be a useful technique, especially for large constructs such as tables as
+     it allows the children to be inferred without needing to be
+     explicitly referenced in the `intent` as would be the case with an <i>applicaton</i>.
+     For example, `&lt;mtable intent=":array">...` might read the table as
```
davidcarlisle, line 353: The examples here should probably align with the core properties (even if not explicitly referencing that list) which (today at least) would be `matrix` and `system-of-equations` rather than `array` and  `aligned-equations`

brucemiller: done; thanks!

davidcarlisle (approved):

src/intent.html
```diff
@@ -182,52 +172,24 @@ <h3 id="mixing_intent_dictionaries">Intent Concept Dictionaries</h3>
     <p>
       [=Intent=] Concept names are maintained in two lists, each maintained in the
       <a href="https://github.com/w3c/mathml-docs">w3c/mathml-docs GitHub repository</a>.
+      Note that while these concept dictionaries are published as HTML tables (based on yaml data),
+      there is no requirement on how a system implements the mapping from concepts to speech hints.
+      Rather than a fixed list or hash table, it might use XPath matching, regular expressions,
+      appropriately  trained generative AI or any other suitable mechansim.
+      The only requirement si that it should accept the cases listed in the Core concept dictionary
```
davidcarlisle:
```suggestion
      The only requirement is that it should accept the cases listed in the Core concept dictionary
```

davidcarlisle (approved):
one trivial typo marked in line (si/is)  but otherwise last round of changes  look good

NSoiffer:
I think this is a very good rewrite. There is more work to do (e.g., add some text on `:literal`).  But this is a good rearrangement of the material and clarifies that Intent Concept Dictionary is a general idea that can be implemented in a wide variety of ways.

NSoiffer (approved):

davidcarlisle:
@brucemiller @NSoiffer  not sure why the pipeline is failing (the spec seems to have built and pushed to gh-pages)

davidcarlisle:
I'm not quite sure what it was saying, but the failure seems to be an artifact of the fork/PR.
Builds all green again now.

PR #516, opened 2024-11-07
minor proofreading edits, 11/07/2024

dginev:
This PR is not related to active action items.
It contains misc edits from proofreading that are not meant to introduce any changes:

- consistent `descendant` spelling
- removed an extra `used by`
- changed `AT's` to `supported`
- added speech part to the example for underscore function use
- (I think?) clearer word order on concept addition/removal 
- dropped dash from [unannotated](https://dictionary.cambridge.org/us/dictionary/english/unannotated)

If some of these are not useful, feel free to cherrypick anything of interest and close here.

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -104,7 +104,7 @@ S                  := [ \t\n\r]*
 
     <dt><dfn id="intent_reference">reference</dfn></dt>
     <dd>
-    An argument [=reference=] such as <code>$name</code> refers to a descendent element
+    An argument [=reference=] such as <code>$name</code> refers to a descendant element
     that has an attribute <code>arg="name"</code>. Unlike <code class="attribute">id</code>
     attributes, <code class="attribute">arg</code> do not have to be
     unique within a document. When searching for a matching element the
@@ -305,7 +305,7 @@ S                  := [ \t\n\r]*
   <section>
    <h3 id="intent_using">Using Intent Concepts and Properties</h3>
    <p>When the <code class="attribute">intent</code> attribute corresponding to a specific node
-     contains a concept component, the AT's [=Intent Concept Dictionary=] should be consulted.
+     contains a concept component, the supported [=Intent Concept Dictionary=] should be consulted.
      The concept name should be normalized
      (<q>`_`</q> (U+00F5) and <q>`.`</q>  (U+002E) to <q>`-`</q> (U+002D)),
      and compared using <a data-cite="INFRA#ascii-case-insensitive">ASCII case-insensitive</a>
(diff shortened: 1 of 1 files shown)


src/intent.html
```diff
@@ -526,6 +526,7 @@ <h3 id="mixing_intent_examples">Intent Examples</h3>
   &lt;mo arg="script">&#x2032;&lt;/mo>
 &lt;/msup>
 </pre>
+<blockquote>x prime<br/>x superscript prime end superscript</blockquote>
```
davidcarlisle, line 529: I think this should just have `x prime`  (which I just checked also matched what mathcat does) I don't see how you can get `x superscript prime` as the intent masks the presentation so you are explicitly stoping the "superscript" being read (if you are using the intent at all) The system may of course have an option to read the underlying markup ignoring intent but that shouldn't be in this "intent examples" section.

src/intent.html
```diff
@@ -305,7 +305,7 @@ <h3 id="mixing_intent_properties">Intent Properties</h3>
   <section>
    <h3 id="intent_using">Using Intent Concepts and Properties</h3>
    <p>When the <code class="attribute">intent</code> attribute corresponding to a specific node
-     contains a concept component, the AT's [=Intent Concept Dictionary=] should be consulted.
+     contains a concept component, the supported [=Intent Concept Dictionary=] should be consulted.
```
davidcarlisle: I think this is better with AT as currently written, otherwise it is not saying what is supporting this. The important point here is that thi sis system (AT) specific support,

davidcarlisle (changes requested):
I commented on a couple of changes don't seem quite right, all the others look good thanks

davidcarlisle:
> If some of these are not useful, feel free to cherrypick anything of interest and close here.
I could do this, thanks but I'll leave it for a bit for others to comment, Probably at the weekend...

src/intent.html
dginev: Right, exactly, when the intent is used it is `x prime`. That's the starting fragment before the `<br/>`.

The second entry is the speech when intent isn't used - I copied it directly from the example right above, there are several that follow that pattern.

src/intent.html
dginev: I thought I heard you say you prefer that word choice during the meeting today when reviewing a change I had made? No matter, I have removed it to avoid distractions.

src/intent.html
davidcarlisle: they do? I'll look again when it's not midnight, as I say wasn't planning on rushing any edits:-)

davidcarlisle:
actually I think you are right and I did say that (but I think I was wrong in the meeting and it's better as it is) but now it's definitely too late to think about wording

davidcarlisle:
@dginev on the readings with and without intent. You are right, not only do the ones above have that style the text above that block of examples (which I guess I wrote, but didn't read) explicitly says that is what they are

> These cases would be distinguished as follows, showing possible readings with and without intent

So obviously your addition here is good.

That said I think the example display isn't clear. If you jump to a single example, and you are tired it really isn't clear that the the two line text example is line 1 with intent and line 2 without. I think we should experiment with modifying the format, perhaps having an explicit intent/nointent comment on each. But that doesn't need to be in this PR

davidcarlisle:
@dginev thanks for adjusting the branch, there is nothing remotely controversial in the remaining edits so merging now

PR #517, opened 2024-11-12
CSS and Intent 

### LINKED ISSUE #441: Intent and style (mathcolor, CSS, others?)
Recent trends in pedagogical enhancements over mathematical notation use a variety of styling primitives (traditionally realized on the web via CSS, or MathML attributes such as mathcolor, mathvariant, mathbackground , mathsize).

A recent medium-agnostic collection of resources in this direction is the [awesome mathematical notation design](https://fredhohman.com/awesome-mathematical-notation-design/) showcase.

Here is a motivating screenshot from the first example paper in the collection:
[Math Augmentation: How Authors Enhance the Readability of Formulas using Novel Visual Design Practices](https://andrewhead.info/assets/pdf/augmented-formulas.pdf).

The screenshot contains a display equation, paragraph and diagram, all targeted at sighted readers, which use different colors for each separate conceptual piece in each modality. For example, the "incoming light" concept is highlighted in the text in purple, as is its `L_i` term in the display equation, as well as the corresponding arrows in the hand-painted diagram.

The question for this issue is - how do such techniques fit with the `intent` paradigm, if at all, so that AT users can also benefit from the additional visual information.

To start with the simplest possible example, consider a red variable x, with two possible MathML expressions:
```
<mi mathcolor="red">x</mi>
<mi style="color: red;">x</mi>
```

Will AT be able to pick up the color information by default? Should it?
Alternatively, should an author be able to make it explicit that such information is important to also convey to AT users?

The first temptation is to reach for `intent="red-x"`, or `intent="_(_red,$x)`, but one could of course also consider providing the information in the accessible description via `intent="x:red-color-variable"`.

Another example would be the use of an `<mrow style="border:solid;">...</mrow>` to draw a rectangle border around a term, presumably as a means of emphasis or to designate a particular notation. In the case where this has a named notation, `intent` can be used as designed. 

In the case of styling emphasis (as in "boxed"), we are probably closer to `intent=_boxed($this)` or `intent=$this:has-rectangular-frame`, but I am quite unsure at the moment. In addition, if a larger sub-expression had a dedicated color (typically contained within an mrow) it is conceivable that AT may want to mark the "start" and "end" of certain colored zones.

I think the group hasn't considered the relationship between the `intent` mechanism and the styling information in the presentation tree yet, so I am opening an issue to start a discussion.

dginev:
Fixes #441 .

This PR adds a single example and some minimal text, mostly to start discussion on what the exact framing should be. Completely open to being fully reworked.

A preview is visible at the fork:
https://dginev.github.io/mathml/spec.html#mixing_intent_examples_css

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -562,7 +562,34 @@ the following example</p>
 <blockquote>bell number of 2</blockquote>
 </div>
 
-    
+<section>
+  <h4 id="mixing_intent_examples_css">CSS and Style</h4>
+
+  <p>CSS customization of MathML is generally not made available to AT and is ignored
+    in accessible readouts.
+  In cases where authors have meaningful stylistic emphases,
+  or stylized constructs with custom names, using an intent attribute is appropriate.
+  For example, color-coding of subexpressions is often helpful in teaching materials:</p>
+<div class="example mathml mmlcore">
+<pre>
+&lt;mn&gt;1&lt;/mn&gt;&lt;mo&gt;+&lt;/mo&gt;
+&lt;mrow style="padding:0.1em;background-color:lightyellow;"
+      intent="highlighted-step($step)"&gt;
+  &lt;mfrac arg="step"&gt;&lt;mn&gt;6&lt;/mn&gt;&lt;mn&gt;2&lt;/mn&gt;&lt;/mfrac&gt;
+&lt;/mrow&gt;
+&lt;mi&gt;x&lt;/mi&gt;
+&lt;mo&gt;=&lt;/mo&gt;
+&lt;mn&gt;1&lt;/mn&gt;&lt;mo&gt;+&lt;/mo&gt;
+&lt;mn style="padding:0.1em;background-color:lightgreen;"
+    intent="highlighted-result(3)"&gt;3&lt;/mn&gt;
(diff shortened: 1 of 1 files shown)


davidcarlisle:
I quite like the example and the introductory text, I wonder if we could avoid suggesting the duplication and use a selector such as

```
<!DOCTYPE html>
<html>
<head>
  <style>
    *[intent*="highlighted-result"] {padding:0.1em;background-color:lightgreen;}
    *[intent*="highlighted-step"] {padding:0.1em;background-color:lightyellow;}
  </style>
</head>
<body>

 
<math>
<mn>1</mn><mo>+</mo>
<mrow intent="highlighted-step($step)">
  <mfrac arg="step"><mn>6</mn><mn>2</mn></mfrac>
</mrow>
<mi>x</mi>
<mo>=</mo>
<mn>1</mn><mo>+</mo>
<mn intent="highlighted-result(3)">3</mn>
<mi>x</mi>
</math>
</body>
</html>
```

although that may stress the current example display in the spec which isn't set up for separate css, in a real document it would save a lot of duplication though.

This is just musing though the example could go in as it is,

NSoiffer:
Very nice!

PR #515, opened 2024-11-05
Matched/unmatched concepts in Intent Dictionaries

dginev:
This is a rebased version of [the original PR](https://github.com/brucemiller/mathml/pull/1), see that link for discussion.

The current branch has an additional `(optionally)` modifier, for extra clarity.

PR #518, opened 2024-11-12
[wording] no restrictions from supported concept definition 

dginev:
Following #515 and the WG meeting on 11/07/2024, we had an offline discussion with @polx on some minimal text that avoids a strict reading of the "supported concept" definition.

This PR contains our proposal.


davidcarlisle (approved):
Personally I don't think this makes things that much clearer, but more than one person thought the existing text wasn't that clear and this addition seems fine to me so happy to add this. I would have just merged as editorial but since it has been the subject of some WG debate I'll just approve and leave time for Neil to review.

NSoiffer (approved):
I agree with David that this adds nothing new. Nonetheless, since two people expressed confusion and David doesn't object, I'll approve this in the hopes that this makes the rec clearer to some and do the merge.

Issue #103, opened 2019-06-14
Extend mpadded attribute syntax to be closer to MathML3


fred-wang:
In w3c/mathml#81, we agreed that mpadded attributes in MathML Core should only use pseudo-unit with the same vertical/horizontal direction.

Can we simplify it any further? Remove/Deprecate it?

Some thoughts:
* MathML discourage authors from using mpadded for tweaking or conveying meaning: https://mathml-refresh.github.io/mathml/chapter3.html#presm.warnfinetuning
* In general, since MathML3 did not clearly specify rendering, implementation are inconsistent and usage is probably unreliable.
* CSS is more appropriate for this kind of space tuning. I imagine if we better define MathML interactions with out-of-flow position w3c/mathml-core#74, margin/border/padding w3c/mathml#9, vertical-align w3c/mathml-core#273, aspect-ratio etc authors could use these alternatives.
* mpadded is probably used a lot, though. We would need to get better statistics and analysis of current use cases.

fred-wang:
These are the possible non-length values implemented by @distler in itex2MML and copied into TeXZilla:
lspace="-100%width" (\mathllap)
lspace="-50%width" (\mathclap)
height="+[LENGTH][UNIT]" (\mathraisebox)
depth="+[LENGTH][UNIT]" (\mathraisebox)
depth="depth" (\mathraisebox)

depth="depth" can basically be replaced with not using (or ignoring) the attribute. I'm not exactly sure about how the other commands work.

fred-wang:
From https://mathml-refresh.github.io/mathml/chapter3.html#presm.mpadded

For any unsigned number X, if Y = X / 100 then

X%pseudo-unit is equivalent to Ypseudo-unit
+X%pseudo-unit is equivalent to +Ypseudo-unit
-X%pseudo-unit => -Ypseudo-unit

So it seems that this syntax is not useful.

fred-wang:
I've rewritten the description of mpadded in MathML core, so that it no longer refers to MatHML3
https://mathml-refresh.github.io/mathml-core/#adjust-space-around-content-mpadded

Unless I'm mistaken, the current syntax and semantic should more or less match what is in the full spec: 
https://mathml-refresh.github.io/mathml/#presm_mpaddedatt
(with namedspace removed and numbers defined from CSS)

NSoiffer:
[...]

PR #519, opened 2024-11-16
align mpadded with core

davidcarlisle:
This PR addresses issue #103 where the WG resolved to make mathML4 match mathml-core syntax

A rendered version may be seen at

https://davidcarlisle.github.io/mathml/#presm_mpaddedatt

or for the change log at

https://davidcarlisle.github.io/mathml/#chg_presm

Basically this drops the relative lengths such as `+3em` as you can use `calc(100%+3em)` in standard CSS length syntax and it drops  `0.5 height` as you can use `50%` in the most common case of using the existing value while setting the same dimension. You can no longer reference the original height while setting a new width, but this is not a common use case.

### DIFF
--- a/src/changes.html
+++ b/src/changes.html
@@ -105,10 +105,22 @@
     <li>The deprecated element <code class="element">none</code> is
     replaced by an empty <code class="element">mrow</code> throughout,
     to match [[MathML-Core]].</li>
-    <li>The element <code class="element">mlabeledtr</code> and the associacted attributes
+    <li>The element <code class="element">mlabeledtr</code> and the associated attributes
     <code class="attribute">side</code> and <code class="attribute">minlabelspacing</code>
     are no longer specified. They are removed from the default schema but valid in
     the <a href="#parsing_rnc_legacy">Legacy Schema</a>.</li>
+
+    <li>To align with MathML-Core, the special extended syntax for  <code class="element">mpadded</code> length attributes 
+      <code>( "+" | "-" )?
+      <em>unsigned-number</em>
+      ( ("%" <em>pseudo-unit</em>?)
+      | <em>pseudo-unit</em>
+      | <em>unit</em>
+      | <em>namedspace</em>
+      )?</code> is no longer supported.
+    Most of the functionality is still available using standard CSS
+    length syntax. See <a href="#mpadded-lengths">Note: mpadded
+    lengths</a>.</li>
    </ul>
 
   </section>
--- a/src/fundamentals.html
+++ b/src/fundamentals.html
@@ -407,14 +407,14 @@
 	</tbody>
       </table>
 
-    <p>In addition, the attributes on <a href="#presm_mpadded">mpadded</a>
-     allow three <dfn>pseudo-units</dfn>, <code>height</code>,
+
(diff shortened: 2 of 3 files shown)


davidcarlisle:
Not done in this PR is to adjust the schema, moving the mpadded-length syntax to the legacy schema.  If the schema at `mathml-schema` is updated  appendix A here will automatically update.

NSoiffer:
It looks pretty good but the last example is bascially a repeat of the previous "calc" example. Can you make the example of what `depth= height` changes into?

davidcarlisle:
@NSoiffer 
> It looks pretty good but the last example is bascially a repeat of the previous "calc" example. Can you make the example of what `depth= height` changes into?

er yes I will look later I didn't change the examples at all other than just remove the syntax that this PR is removing and replace with `calc`. In particular the fancy boxed images, but perhaps now the syntax is simpler with less custom variants we can lose one of the examples, or show it differently.

davidcarlisle:
@NSoiffer oh sorry I misread your comment
@NSoiffer 
> Can you make the example of what depth= height changes into?

I think that changes in to nothing, I don't think that is supported. Something had to give. You can replace a pseudo-unit that refers to the current dimension by a percentage, but I don't think there is a way to refer to one of the other dimensions

davidcarlisle:
I wrote

> I don't think that is supported. 

You can of course use JavaScript, but it turns out that CSS does support referencing the width when setting height in one specific case, percentage values for padding are based on the width. As this is padding you need to hide the natural height if you want to set the height explicitly.

This more or less works in Chrome/Edge to get a squarish box with height set to the width (I only approximated hiding the natural height here by subtracting 1ex.)

```
<!DOCTYPE html>
<html>
  <head>
   <style>
    .square {padding-top:calc(100% - 1ex)}
    .box {border: solid thin black}
   </style>
  </head>
  <body>
   <math>
    <mi>x</mi>
    <mo>+</mo>
      <menclose class="box" notation="box">
	<mi class="square">abc</mi>
      </menclose>
    <mo>+</mo>
      <menclose class="box" notation="box">
	<mi class="square">abcdef</mi>
      </menclose>
    </math>
  </body>
</html>
```

image


The syntax here is of course weird but it does indicate that setting the height based on width isn't totally alien to the CSS model and perhaps CSS can have a more natural syntax for this at some point.

NSoiffer:
Do you want to add that as an example or do you think it is too big/complicated?

davidcarlisle:
@NSoiffer 
> Do you want to add that as an example or do you think it is too big/complicated?

No, it's too much of a hack (and pads by twice as much in firefox as chrome), and while it's easy to add top or bottom _padding_ specifying the total height this way needs tricks with absolute positioning that takes the content out of the flow and put it back, and more than likely to conflict with math layout somewhere.

I mention it though as it shows that the basic principle of accessing the original width while setting the height isn't totally alien to css. Also while it's not something I think we could put in the spec it may possibly be useful in some kind of polyfill especially if that's needed in a non-javascript context where you only have css.

NSoiffer:
Ok. I'll merge it.

davidcarlisle:
I'll bring the schemas in to line...

PR #520, opened 2024-11-20
Add guidance for using aria-description

### LINKED ISSUE #493: Clarify property use as descriptions
The spec text currently states:
https://w3c.github.io/mathml/#intent_property
> property
>    A [property](https://w3c.github.io/mathml/#intent_property) annotates the intent with an additional property which may be used by the system to adjust the generated speech or Braille in system specifc ways. The property may be directly related to the speech form, such as :infix or indirectly affect the style of speech with properties such as :unit or :chemistry
>
>    The list of properties supported by any system is open but should include the core properties as described below.

It seems important to extend this text to match the original `is-a` motivation (#426). The text from ARIA [aria-description](https://w3c.github.io/aria/#aria-description) and the definition of [Accessible Description](https://www.w3.org/TR/accname-1.2/#dfn-accessible-description) can be a good inspiration. As a first stab:

> Non-core properties can be treated as a string value that describes or annotates the intent head, or if the head is empty - the current element. Non-core properties can provide more verbose information, as necessary, to offer an accessible description.

Suggestions welcome.

dginev:
Closes #493 in favour of `aria-description`.

This PR was an action item from the WG meeting on Nov 14, 2024. The text and example are completely tentative, rewordings welcome.

Preview at:
https://dginev.github.io/mathml/spec.html#acc_descriptions

### DIFF
--- a/src/accessibility.html
+++ b/src/accessibility.html
@@ -424,6 +424,38 @@ could be marked up as an operator by means of
 
 </section>
 
+<section>
+  <h5  id="acc_descriptions">Accessible Descriptions</h5>
+
+<p>It is sometimes beneficial to a reader to have additional, auxiliary information for a given
+   mathematical construct. This is particularly the case in educational materials where
+   newly introduced syntax benefits from repeated reinforcement. Such information is often
+   too verbose for more familiar readers, indeed even to the same person after their first reading.
+   Hence, common AT behavior has been to only vocalize an
+   <a href="https://www.w3.org/TR/accname-1.2/#dfn-accessible-description">accessible description</a>
+   on user request, omitting it by default.</p>
+<p>It is appropriate to provide such descriptions using the ARIA 1.3 attribute 
+  <a href="https://www.w3.org/TR/wai-aria-1.3/#aria-description"><code class="attribute">aria-description</code></a>,
+  which allows for a literal string value to annotate its host element.
+  As an example, consider the minimal markup for the circumference formula,
+  with each non-trivial component described.
+</p>
+
+<div class="example mathml mmlcore">
+<pre>
+&lt;mrow aria-description="circumference of a circle"&gt;
+  &lt;mn&gt;2&lt;/mn&gt;
+  &lt;mi aria-description="mathematical constant"&gt;π&lt;/mi&gt;
+  &lt;mi aria-description="radius variable"&gt;r&lt;/mi&gt;
+&lt;/mrow&gt;
+</pre>
+</div>
+
+<p>While <code class="attribute">aria-description</code> has been used on its own for brevity,
+  it is recommended to use it together with an <code class="attribute">intent</code> annotation,
+  as appropriate.
+</p>
+</section>
    </section>
   </section>
  </section>
--- a/src/accessibility.html
+++ b/src/accessibility.html
@@ -455,6 +455,12 @@ could be marked up as an operator by means of
   it is recommended to use it together with an <code class="attribute">intent</code> annotation,
   as appropriate.
 </p>
+<p> To ensure equal access, when <code class="attribute">aria-description</code>
+  or <code class="attribute">intent</code> are used,
+  useful descriptions should also be made visible on the page.
+  A common affordance to achieve that is interactively displaying
+  a tooltip containing the description.
+</p>
 </section>
    </section>
   </section>


NSoiffer (commented):
The text looks good. My only concern, which I don't have a solution for other than maybe using in conjunction with something that provides a tooltip, is a reverse "equal access". A goal of accessibility it to provide equal access to material for people with different impairments. In this case, the description is _only_ provided to people using AT. Unless the description is something that is provided to the sighted person in the content in another manner (such as an arrow what the description is such as the radius of a circle), then it isn't equal access.

NSoiffer:
@Steve-Noble: you wrote a lot of the text here. Can you take a look at Deyan's suggestion and see if you think this is appropriate?

dginev:
> Unless the description is something that is provided to the sighted person in the content in another manner (such as an arrow what the description is such as the radius of a circle), then it isn't equal access.

That is a fair requirement. But I think the Web Platform is a good vehicle to achieve this, right? 
CSS+JS can leverage `math *[aria-description]` selectors to scaffold a (non-standard) tooltip component. 

Here is a very primitive example:
https://codepen.io/dginev/pen/OJKYKwN

It occurs to me that `intent` expressions would also appear to be subject to the equal access requirement. So maybe a shared solution is possible.

NSoiffer:
Maybe add a sentence or so that says something like 'if aria-description is used, then some technique such as having a tool tip with similar information should be provided for non-screen reader users'.

As I think about this more, it seems backward. Although I was the one to suggest adding this to the appendix, maybe the main reference should be in the [maction](https://w3c.github.io/mathml/#presm_enliven) section. It could get mentioned as means of conveying information to AT that might otherwise only be visually accessible. There would still need to be a brief mention of `aria-description` in this section. On the other hand, the maction section currently doesn't mention any tooltip-like actions, so maybe mention of using `aria-description` doesn't make sense there.

An aside: I briefly wondered if the title attribute made sense to include in MathML and quickly found this:
> "If you want to hide content from mobile and tablet users as well as assistive tech users and keyboard only users, use the title attribute." — [The Paciello Group blog](https://www.paciellogroup.com/blog/2013/01/using-the-html-title-attribute-updated/)

This is from a page MDN references:  [Tooltips & Toggletips](https://inclusive-components.design/tooltips-toggletips/). That has some brief mentions of `aria-describedby` usages.

Steve-Noble:
I would tend to echo Neil's comment. Information that essentially summarizes a mathematical expression should have an equal access method beyond only making it available to a screen reader. I think the general principle is that if the description would also be useful to sighted users, it's better to make it visible on the page in some way and use aria-describedby to reference it instead. 

"Since aria-description is not visible to users, it’s sometimes used to give an element a special accessible description for screen reader users only. Consider if the accessible description provided in the aria-description attribute might be useful to any sighted users. If it is, make the description visible on the page and instead use aria-describedby to reference it.

See https://govtnz.github.io/web-a11y-guidance/ka/fundamental-concepts-in-web-a11y/accessible-names-and-descriptions/accessible-descriptions.html#be-careful-with-aria-description

dginev:
@Steve-Noble and @NSoiffer thank you for the feedback.

I have attempted to do a 3-way synthesis with some extra text in 8a19efbc7cd40d5f9c24df26dc92913762d49a7c, now also in the [preview site](https://dginev.github.io/mathml/spec.html#acc_descriptions).

I have tried to stay away from discussing `aria-describedby` for now, since I am not sure how to best formulate that _some_ ATs may be limited to only a single `<math>` element. I think the Web Platform allows for full-document id lookup. It even makes me wonder if MathML Core would want to describe ARIA interplay with different scope than MathML Full would. So I'm deferring on that for now.

NSoiffer:
@Steve-Noble: the problem with `aria-described-by`, at least for implementations that use MathCAT, is that the `id` will be outside of the MathML and so MathCAT can't see it and can't present it to screen reader users. At least that's the case with the current MathCAT. Potentially maybe a new callback function or DOM pointer can be introduced in a new call, but I think that's a lot of complication. It's the reason why @dginev went with `aria-description`.

NSoiffer:
I like the new text.

NSoiffer:
Based on the conversation at today's meeting where no one objected to the text, I'm approving this.

Issue #178, opened 2019-12-13
Make MathML attributes ASCII case-insensitive


fred-wang:
This is a follow-up of #22 ; we decided to follow HTML/CSS which treat things as ASCII case-insensitive.  Concretely, ftp://ftp.unicode.org/Public/UNIDATA/CaseFolding.txt has this line:

    017F; C; 0073; # LATIN SMALL LETTER LONG S

which means that `falſe` is case-insensitively equal to `false`. However, it is not ASCII case-insensitively equal to `false` (only a-z <-> A-Z equivalence are considered in that case).

Currently, the MathML Core spec just says "case-insensitive". 

Note: for CSS colors, I reported https://github.com/w3c/csswg-drafts/issues/4599

fred-wang:
I did a quick check and for the MathML-specific definitions, I only see case-insensitive against strings with ASCII letters and dashes. So the only difference would be for "LATIN SMALL LETTER LONG S", "KELVIN SIGN" and maybe a few "LATIN SMALL LIGATURE" (e.g.  `double-STruck`). Unlikely for "LATIN CAPITAL LETTER I WITH DOT ABOVE" if the Turkish rule is used. See https://github.com/w3c/csswg-drafts/issues/4599#issuecomment-565794132

NSoiffer:
That's a good catch. I'm pretty sure we all agree that we only mean ASCII case-insensitivity. I suggest we add the following to the spec, which is a slight rewording from the [HTML spec](https://html.spec.whatwg.org/multipage/syntax.html#writing):

> Many strings in the HTML and CSS syntax (e.g. the names of elements and their attributes) are case-insensitive, but only for ASCII upper alphas and ASCII lower alphas. For convenience, in this specification this is just referred to as "case-insensitive".

I suggest this goes into [Appendix G.1: Document Conventions](https://mathml-refresh.github.io/mathml-core/#document-conventions).

fred-wang:
I would prefer to be explicit everywhere and use "ASCII case-insensitive" with a link to https://infra.spec.whatwg.org/#ascii-case-insensitive ; this seems to be what the HTML and CSS specifications do (or how they would be fixed it e.g. https://github.com/w3c/csswg-drafts/issues/4599#issuecomment-565599403). I'm sure if we just keep case-insensitive as it is now, people will easily not read the appendix. We should also avoid duplicating definition from HTML5 as it was mentioned in another issue.

fred-wang:
[...]

PR #521, opened 2024-11-24
defined boolean attribute type

davidcarlisle:
Add a boolean attrbute type (refencing the existing definition in MathML-Core), replace all `"true" | "false"` by a reference to boolean.

This addresses a main case highlighted by #178

### DIFF
--- a/src/fundamentals.html
+++ b/src/fundamentals.html
@@ -158,39 +158,39 @@
      <tbody>
 
       <tr>
-       <td id="type_boolean"><dfn><em>boolean</em></dfn></td>
+       <td id="type_boolean"><dfn class="no-export"><em>boolean</em></dfn></td>
        <td>As defined in [[MathML-Core]], a string that is an
               ASCII case-insensitive match to <code>true</code> or
               <code>false</code>.</td>
       </tr>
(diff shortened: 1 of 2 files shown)


NSoiffer:
Nothing wrong with this as a start, but don't most of the attribute values need to be ASCII case insensitive? Or do you think only those that use boolean values?

davidcarlisle:
@NSoiffer  I think values that are case insensitive in mathml-core need to be case insensitive in full, so that anything valid in core is valid in full (and that's already what the schema implemented even though the spec wording did not). Making other things case insensitive is possibly the right thing to do but less important I think. Certainly in an xml world there is less pressure to be case insensitive, eg xml schema for xhtml and svg are typically case sensitive, it is the versions embedded in html where this is most natural.

But I was not suggesting that we don't change the others, just doing a simple but important case as low hanging fruit, that could easily be done and reviewed on a wet afternoon.

NSoiffer:
Yes, very sensible. I'll open an issue on whether we should make all the fixed value ones (such as "left"/"center"/"right") be case sensitive or not. If we decide not to, then there's nothing to do.

PR #529, opened 2025-03-13
remove open properties list (WG call 2025-03-13)

davidcarlisle:
As agreed on the WG call of 2025-03-13, the consensus is to drop the open properties list.

This PR adjusts the wording in the spec to just refer to a single properties list.

It is dependent on two PR in mathml-docs

add `:equation-label`

https://github.com/w3c/mathml-docs/pull/84

move `:pause-xxx` to the main list.

https://github.com/w3c/mathml-docs/pull/86

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -221,20 +221,14 @@ S                  := [ \t\n\r]*
    but as with any properties, is by default ignored but may have a
    system-specific effect.</p>
 
-    <p>As with [=Concepts=], The Working group maintains two lists
-      of [=property=] values.</p>
+    <p>Similar to the lists of [=Concepts=], The Working Group maintains a list of [=property=] values that should be implemented.</p>
     <ul>
-      <li><dfn id="intent-core-properties">Core properties</dfn>: This is a
-      <a href="https://w3c.github.io/mathml-docs/intent-core-properties">list core of properties</a>
+      <li><dfn id="intent-core-properties">Properties list</dfn>: This is the full
+      <a href="https://w3c.github.io/mathml-docs/intent-core-properties">list of properties</a>
       maintained by the Math Working Group</li>
-       <li><dfn id="intent-open-properties">Open properties</dfn>: This is an
-      <a href="https://w3c.github.io/mathml-docs/intent-open-properties">open list of properties</a>
-      with contributions welcome from the community.
-      Implementors of MathML systems that implement additional properties are encouraged
-      to make a pull request to add them to the list of  [=Open Properties=].</li>
     </ul>
-   <p>The definitive list of [=Core Properties=] is maintained at
-   Github. Here, we describe the major classes of property affecting speech generation
+   <p>The definitive  [=Properties list=] is maintained at
+   GitHub. Here, we describe the major classes of property affecting speech generation
    below.</p>
    <dl>
     <dt id="intent_fixity_hint"><code>:prefix</code>,
--- a/src/intent.html
+++ b/src/intent.html
@@ -221,15 +221,13 @@ S                  := [ \t\n\r]*
    but as with any properties, is by default ignored but may have a
    system-specific effect.</p>
 
-    <p>Similar to the lists of [=Concepts=], The Working Group maintains a list of [=property=] values that should be implemented.</p>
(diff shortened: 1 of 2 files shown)


dginev (approved):

src/intent.html
```diff
@@ -221,20 +221,14 @@ <h3 id="mixing_intent_properties">Intent Properties</h3>
...
       maintained by the Math Working Group</li>
-       <li><dfn id="intent-open-properties">Open properties</dfn>: This is an
-      <a href="https://w3c.github.io/mathml-docs/intent-open-properties">open list of properties</a>
-      with contributions welcome from the community.
-      Implementors of MathML systems that implement additional properties are encouraged
-      to make a pull request to add them to the list of  [=Open Properties=].</li>
     </ul>
-   <p>The definitive list of [=Core Properties=] is maintained at
-   Github. Here, we describe the major classes of property affecting speech generation
+   <p>The definitive  [=Properties list=] is maintained at
```
physikerwelt: too much space

src/intent.html
```diff
@@ -221,20 +221,14 @@ <h3 id="mixing_intent_properties">Intent Properties</h3>
    but as with any properties, is by default ignored but may have a
    system-specific effect.</p>
 
-    <p>As with [=Concepts=], The Working group maintains two lists
-      of [=property=] values.</p>
+    <p>Similar to the lists of [=Concepts=], The Working Group maintains a list of [=property=] values that should be implemented.</p>
     <ul>
-      <li><dfn id="intent-core-properties">Core properties</dfn>: This is a
-      <a href="https://w3c.github.io/mathml-docs/intent-core-properties">list core of properties</a>
+      <li><dfn id="intent-core-properties">Properties list</dfn>: This is the full
```
physikerwelt: what does `full` mean

physikerwelt: The id could be updated (while this would break permanent links)

physikerwelt:
I support this

src/intent.html
davidcarlisle: well it;s html so the space does no harm, but I cleanred up (and a spurious "below" was left in the sentence from the original form (checkin to follow)

src/intent.html
davidcarlisle: it was the wording used before, the full list is maintained in the yaml and an "edited highlights" appears here. Actually if the full list doesn't get much bigger  it may make sense to have it all here, but for now I'll merge this with the following paragraph and re-word. The version here has too much text left over from describing two lists

davidcarlisle: yes I left this, the correspondence with the core concept list is till there even though we decided (for whatever reason) to drop the open list for properties.

dginev: > we decided (for whatever reason) 

The reason was removing process friction in deciding whether new additions belong in Core or Open, as it was hard to have clear consensus on the bar for entering either list.

Clarifying, since it appeared necessary.

davidcarlisle:
> I support this

@physikerwelt 
thanks for the review, I've cleaned up the text (I hope) if you are OK with me merging can you approve with the tick just for the record, then with two ticks (one from Deyan already) I'll do the merges and we can look how it all looks "in place",

src/intent.html
davidcarlisle: certainly we are doing it to remove process friction, I don't think anyone understood why you think it does, but as I say, if it does and we can move on, then that's all good. Since the open list (should have been) empty anyway it's existence shouldn't have been contentious, but conversely its loss isn't any great loss.

dginev: > I don't think anyone understood why you think it does

That's rather uncharitable and unneeded, I take your point.

davidcarlisle: @dginev I can delete the comment if you wish, but  I'm not sure why you say it is uncharitable, I'd say it was  fairly accurate sentiment of the meeting. the open properies list is being removed not on any technical grounds but simply because you said removing it would remove the process block that meant we could not progress on the spec without spending weeks discussing each property and revisiting 2 year old decisions. As has been discussed, the list should not have contained any property that was being used in public documents so removing it is not really any loss, so definitely a price worth paying.

physikerwelt (approved):

davidcarlisle:
This is failing a spurious validation error, so I can't merge in the GUI here, I'll merge locally

Issue #284, opened 2022-03-10
Make the sample presentation of Strict Content use intent


NSoiffer:
There are currently "sample" renderings. Let's make this use `intent`.

davidcarlisle:
Randomly picking a simple example

https://w3c.github.io/mathml/#contm_cn_extended

has

```
<cn type="rational">22<sep/>7</cn>
```

rendering as

```
<mrow><mn>22</mn><mo>/</mo><mn>7</mn></mrow>
```

My OpenMath/ContentMathML bias wants me to suggest


```
<mrow intent="rational-number($a,$b)"><mn arg="a">22</mn><mo>/</mo><mn arg="b">7</mn></mrow>
```

But that's not in the core intent list and I don't think I'd read an element of ℚ differently to the operation of dividing two integers, so I guess either of the core intent entries `divide` or `fraction` is suitable (I'd prefer `fraction` if forced to choose). Or we'd say this is the default intent for an infix  `/` so don't add `intent` here?

dginev:
with #426 in mind that example reads closer to:
```xml
<mrow intent-isa="rational-number"><mn>22</mn><mo intent="division">/</mo><mn>7</mn></mrow>
```
the associated concepts from wiki are:
- [division](https://en.wikipedia.org/wiki/Division_(mathematics))
- [rational-number](https://en.wikipedia.org/wiki/Rational_number)

For your "sample presentation" examples you likely don't need to restrict yourself to intent values *only* from the Core list, as that list is now in the process of being compressed to a point where no realistic text could be annotated using only entries from it. To illustrate, we've discussed that [absolute-value](https://en.wikipedia.org/wiki/Absolute_value) is likely to not be in the Core list, but it will be needed for the sample presentation of `<abs>`, in the [4.3.7.2.1 Examples](https://w3c.github.io/mathml/#integer1.unary.ex1), namely:
```xml
<apply><abs/><ci>x</ci></apply>
```
```xml
<mrow intent="absolute-value($x)">
  <mo>|</mo>
  <mi arg="x">x</mi>
  <mo>|</mo>
</mrow>
```

davidcarlisle:
Consensus is to allow the `intent` attribute to  be used in some of the sample presentation examples, but we do not have  the resources to commit to regenerating every example, so these will be added on a case by case basis where it seems appropriate.

davidcarlisle:
The above commit adds intent to some examples, showing the principle that intent could be used, I suggest closing this issue.

[...]

PR #530, opened 2025-03-15
add intent to selected examples for #284

davidcarlisle:
as discussed in #284 not adding this everywhere but using a few core intent concepts where it seems reasonable, including the canonical examples of |x| used for absolute value and cardinality and overbar used for complex conjugate


NSoiffer (approved):
Look ok.

Issue #388, opened 2022-06-19
scriptlevel and displaystyle


davidcarlisle:
https://w3c.github.io/mathml/#presm_scriptlevel

says of scriptlevel and displaystyle attributes:

> They also may be set explicitly via the displaystyle and scriptlevel attributes on the [mstyle](https://w3c.github.io/mathml/#presm_mstyle) element or the displaystyle attribute of [mtable](https://w3c.github.io/mathml/#presm_mtable). In all other cases, they are inherited from the node's parent.

similarly the schema "adds" these  to `<mstyle>`

https://w3c.github.io/mathml/#parsing_mstyle.specificattributes

and mstyle attribute table lists them:

https://w3c.github.io/mathml/#presm_mstyle_attrs

However all these descriptions come from Mathml3.

Core makes them global

https://w3c.github.io/mathml-core/#global-attributes

so in Full  they are common to all Presentation elements.

https://w3c.github.io/mathml/#presm_presatt

So should not be described as specific to mstyle.

----

This is assuming we are happy to relax the use of `scriptlevel` in this way. (I do not recall why it was restricted in earlier versions).

If instead we want to keep the restriction, then the most natural place to do that would be to change Core so these are not global attributes there, but keeping Core as it is, and changing Full to match looks OK to me.

NSoiffer:
I think the reason to avoid them on all elements was because that we didn't need to think about how they interreacted with an element that changed the scriptlevel. Putting all that logic on/isolating it to mstyle seemed easiest for the spec and implementations.

It seems safe to change them to be global, but we need to say that the new value overrides the existing value on the current element, not on the any children that change it (they make use of the new value as part of their changes).

davidcarlisle:
PR is merged, closing here.

PR #531, opened 2025-03-15
global scriptlevel for #388

davidcarlisle:
As discussed in #388 Core makes `scriptlevel` global and this adjusts mathml4 to match.

The wording is maybe a little awkward but first thing is to get it adjusted.

this adjusts the wording in the **3.1.6 Displaystyle and Scriptlevel** section and deletes the two attributes from the table of `mtable` attributes (as they are now shared by all presentation)

This PR doesn't adjust the schema (which are in the separate mathml-schema repo) but that was already allowing these everywhere as the presentation mathml schema imports the core one, which defines them as global.

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -4,7 +4,6 @@
 # https://w3c.github.io/spec-prod/#examples
 name: CI
 on:
-  pull_request: {}
   push:
     branches: [main]
 jobs:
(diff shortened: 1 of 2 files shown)


NSoiffer (commented):
"if" -> "is (2x) in the following:
Note that if <code class="attribute">scriptlevel</code> if used with a `+` or `-` sign then the effective `scriptlevel` is incremented or decremented by the value.
   If  <code class="attribute">scriptlevel</code> if used with an unsigned integer, the effective `scriptlevel` is set to that value.

davidcarlisle:
> "if" -> "is (2x) in the following: Note that if `scriptlevel` if used with a `+` or `-` sign then the effective `scriptlevel` is incremented or decremented by the value. If `scriptlevel` if used with an unsigned integer, the effective `scriptlevel` is set to that value.

oops fixed at [82484dd](https://github.com/w3c/mathml/pull/531/commits/82484dd21b33d0445b349145b910ffae19c7eb47)

PR #532, opened 2025-04-03
Describe maligngroup as a transfromation to right-left aligned mtd pairs

davidcarlisle:
This is a draft PR as I'm not happy with the wording yet but generating a PR to make it easy for people to see the diff and track or comment on progress.

A rendered view of the new text is available at

https://davidcarlisle.github.io/mathml/spec.html#presm_malign


NSoiffer:
This should be deleted:
>Note:
>Can malignmark elements occur inside of tokens? 

**done**

---

> The above rules are sufficient to explain the MathML representation of the example given near the start of [this section](https://davidcarlisle.github.io/mathml/#presm_malign).

There is no example there anymore; it is after the code. It probably should be moved to above the code (and the text changed).

**done**

---
Delete section "3.5.4.6 A simple alignment algorithm"

My inclination is to add a warning that maligngroup/malignmark may go away in future versions of MathML. But that would need to be discussed by the group.

davidcarlisle:
@NSoiffer  as agreed in today's call  last remaining attribute removed and the example simplified to just align on the variables and = and not simultaneously decimal align the coefficients. I think this is could be merged now to make review of the whole doc easier.

NSoiffer (changes requested):
3.5.4.1 Description: 
"that can be used to vertically align" => "that can be used to horizontally align"  (e.g., eft/right alignment are horizontal, not vertical)

"each equation, in the example above" -- the example is now gone. Maybe the example at the end of the section?

"no element within the table cell's content should occur entirely before its first maligngroup element": that seems a bit awkward/overly complicated. This implies that even an mi element can't be before the first maligngroup. Wouldn't it be better to say that if an maligngroup in a table cell, the first leaf element in the cell must be maligngroup? You kind of say this a bit further down in the section.

"Expressions in a column that are to have no alignment groups should contain no maligngroup elements." => "contain no maligngroup elements and no malgnmark elements". 

"Expressions with no alignment groups are aligned using only the columnalign attribute that applies to the table column as a whole."  Mention CSS alignment options?

You updated the MathML code example to get rid of the decimal pts, but forgot to update the displayed version.

After fixing the above, feel free to push to the main repo.

davidcarlisle:
> 3.5.4.1 Description: "that can be used to vertically align" => "that can be used to horizontally align" (e.g., eft/right alignment are horizontal, not vertical)
> 

Changed although that text has been there since MathML 1 last century:-)

https://www.w3.org/1999/07/REC-MathML-19990707/chap3_5.html#sec3.5.4


> "each equation, in the example above" -- the example is now gone. Maybe the example at the end of the section?
> 

done

> "no element within the table cell's content should occur entirely before its first maligngroup element": that seems a bit awkward/overly complicated. This implies that even an mi element can't be before the first maligngroup. Wouldn't it be better to say that if an maligngroup in a table cell, the first leaf element in the cell must be maligngroup? You kind of say this a bit further down in the section

this text also from MathML 1. I did try to simplify it before but hard to do in a compatible way the "entirely within" phrasing 
is a bit weird but also used in the previous sentence to describe the split between groups, so using it for this restriction on teh first group is OK I think?

> "Expressions in a column that are to have no alignment groups should contain no maligngroup elements." => "contain no maligngroup elements and no malgnmark elements".

I did change this (although perhaps should change back?). The original text only restricing `maligngroup` was intentional. `malignmark` that appear outside of an alignment group are not an error they are basically `<mspace width="0pt"/>` as explained in the first para of the following section.

> 
> "Expressions with no alignment groups are aligned using only the columnalign attribute that applies to the table column as a whole." Mention CSS alignment options?

added

> 
> You updated the MathML code example to get rid of the decimal pts, but forgot to update the displayed version.

done

> 
> After fixing the above, feel free to push to the main repo.

Issue #385, opened 2022-06-13
comments on chapter 3


davidcarlisle:
I was doing some pubrules and other sanity checking on the spec and collected a few comments, thought I'd try a "combo" issue rather than lots of small PR.  I added checkbox on main items that can be checked once considered

First a note that you can now use `?open`  eg https://w3c.github.io/mathml/?open  to view the spec with all folded sections unfolded:  makes it much easier to review the examples.

* [ ] suboptimal polyfills.

  It's not that surprising that not everything covered by polyfills but in the respec editor's draft where images are   not shown you get a misleading displayed rendering, with no warning. We could (perhaps?) show the images always or at least in the new `?open` mode for reviewing examples.  In `spec.html` you only see the image, not the polyfill so it is not an issue there.  I made the screenshots here from a local version showing both image and polyfill

    * linebreaking (no breaks) https://w3c.github.io/mathml/?open#presm_mo_linebreaks
    
      ![image](https://user-images.githubusercontent.com/1268738/173344461-70c3106d-1826-47bc-a163-21149b038dcb.png)

    * menclose no sloping line in phasorangle https://w3c.github.io/mathml/index.html?open#examples-11
       
      ![image](https://user-images.githubusercontent.com/1268738/173344334-05ac3d62-75d3-4f74-9af3-f7b86ad09fab.png)

   * menclose non-cursive madruwb  https://w3c.github.io/mathml/index.html?open#examples-11

     ![image](https://user-images.githubusercontent.com/1268738/173344637-d0a73df2-9e0f-40f2-b741-8b0eea7df3ec.png)

  * malign not aligned https://w3c.github.io/mathml/index.html?open#mathml-representation-of-an-alignment-example
  
    ![image](https://user-images.githubusercontent.com/1268738/173345558-f49c52be-36ba-4cd4-9c9d-d61d0b335b58.png)

   *  `-`  in mstack becoming hyphen (as it ends up in an html table) http://localhost:8080/w3c/mathml/index.html?open#presm_addsub

      ![image](https://user-images.githubusercontent.com/1268738/173346112-2f47ef55-8a2e-44c2-b317-db381382fbed.png)

      __FIXED__: 9/25

* [x] Remaining Numeric Character ref  `<mmultiscripts><mo>&#x0644;</mo>`  must have missed my global update for some reason. (This seems correct as is: no change)

 * [x] Wide examples needing scrollbar https://w3c.github.io/mathml/?open#presm_mo_linebreaks
[...]

PR #540, opened 2025-09-16
delete bold plus example for #385

davidcarlisle:
as described in #385 using `mathvariant='bold'` with `+` is incorrect, this simply deletes the example. It also removes the  white space at start and end of the `<mo>` in the remaining examples to be consistent with the current definition in MathML-Core which does not strip this space.


NSoiffer:
Makes sense about "bold". Also deleting the spaces is good given the current(?) lack of white space trimming in Chrome.

Issue #524, opened 2025-02-06
Naming the arguments to intent: new property needed?


NSoiffer:
When navigating 2D structures with AT, it is very common to name the part of the expression one has navigated to. For example, if you have a fraction, then when you "zoom into" the fraction, you will often hear "in numerator" and/or "end numerator" when leaving it. Same for the denominator. For "msup", "base" and "superscript" (or "exponent") might be used. If an intent is given that is not in core, there is no way for AT to use a concept-specific name when navigating.

As an example of how one could handle this, one could add something like ":navname-xxx" to the arguments. As an example
```
<math>
    <mrow intent="fraction-like($x, _over_, $y)">
        <mo>\</mo>
        <mtable>
            <mtr><mtd><mi arg='x' intent=':navname-num'>a</mi></mtd></mtr>
            <mtr><mtd><mi arg='y' intent=':navname-denom'>b</mi></mtd></mtr>
        </mtable>
        <mo>/</mo>
    </mrow>
</math>
```

In this example, the user might hear "fraction-like of a and b end fraction-like" when listening to the expression. If they decide to navigate into it, they might hear "in num, x". Moving right, they might hear "out of num, in denom, y". Without the names, they might hear some generic phrases: "in arg 1" and "out of arg 1, in arg 2 y". That's not ideal, especially when you have nested expressions so that there several "arg 1"s, etc.

We at one point discussed a `param=value` for properties but felt that was too complex. This is a "data-xxx"-like ugly version of it where AT would could recognize `navname-` as being important if it is doing navigation and get the name to use by choosing the string that follows it. The main benefit to this approach over the param-value approach is that this doesn't introduce new syntax to the grammar for intent.

dginev:
> When navigating 2D structures with AT

Introducing dedicated markup extensions for "navigation" is an interesting topic, but likely out of scope for the current charter, given the date.

> If an intent is given that is not in core, there is no way for AT to use a concept-specific name when navigating.

This is incorrectly stated. Any "supported concept" for an AT can have enhanced treatment, including Open concepts. It is rather the "unknown concepts" (to an AT) which have no expectation to use any enhanced techniques when navigating into the arguments. I think that is a healthy state.

[...]

PR #541, opened 2025-09-18
list pause and use-argname properties

davidcarlisle:
following #524  updating the abbreviated list of core properties in the spec

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -294,7 +294,13 @@ S                  := [ \t\n\r]*
       <li><code>:equation-label</code> may be used on an <code>mtd</code> to mark a cell containing an equation label.</li>
      </ul>
     </dd>
+    <dt id="intent_pause_properties"><code>:pause-short</code>, <code>:pause-medium</code>, <code>:pause-long</code>, <code>:pause-xlong</code></dt>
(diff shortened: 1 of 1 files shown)


NSoiffer:
I'll merge this, but we need to define "self-property" somewhere before its use.

davidcarlisle:
@NSoiffer 

that's odd the build failed again, it did with the last one you merged too, but I just forced a rebuild by touching the top level index and it re-appeared. the gh-pages spec is 404 at the moment (the one in /TR space is OK)

Too late to worry about that tonight, but on "self-property" I thought it followed from the grammar but we could make an explicit defined term by changing

> 
> **self-property-list**
> 
>  At the top level, an [intent](https://www.w3.org/TR/mathml4/#dfn-intent) may consist of just a non-empty list of properties. These apply to the current element as described in [5.4 Using Intent Concepts and Properties](https://www.w3.org/TR/mathml4/#intent_using).

To

**self-property-list**

At the top level, an [intent](https://www.w3.org/TR/mathml4/#dfn-intent) may consist of just a non-empty list of properties. These apply to the current element as described in [5.4 Using Intent Concepts and Properties](https://www.w3.org/TR/mathml4/#intent_using). Properties used in this way are known as _self-properties_.

PR #543, opened 2025-10-11
Intent interpretation

davidcarlisle:
This PR adjusts the grammar to allow properties on applications and adds descriptions of interpretation of intent arg references and self-properties as suggested in #449

A rendered version is available at

https://davidcarlisle.github.io/mathml/spec.html#mixing_intent_grammar

### DIFF
--- a/src/intent.html
+++ b/src/intent.html
@@ -61,7 +61,7 @@
 intent             := self-property-list | expression
 self-property-list := property+ S    
 expression         := S ( term property* ) S 
-term               := concept-or-literal | number | reference | application
+<dfn id="intent_term">term</dfn>               := concept-or-literal | number | reference | application
 concept-or-literal := NCName
 number             := '-'? \d+ ( '.' \d+ )?
 reference          := '$' NCName
@@ -140,8 +140,7 @@ S                  := [ \t\n\r]*
      the core properties as described below.</p>
     </dd>
 
-<!--    <dt><dfn id="intent_property_list">self-property-list</dfn></dt>-->
-    <dt>self-property-list</dt>
+    <dt><dfn id="intent_property_list">self-property-list</dfn></dt>
     <dd>At the top level, an [=intent=] may consist of just a
     non-empty list of properties. These apply to the current element
     as described in <a href="#intent_using"></a>.</dd>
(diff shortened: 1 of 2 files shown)


NSoiffer (commented):
Some suggested edits:
* ...system dependent however systems... => system dependent. However, systems
* If multiple properties appear in a [self-property-list](...) or [term](...) then they  => add a comma before "then"
* "_parameter_" is italicized, but there is no definition of what a parameter is. I think somewhere there should be a statement that some properties can have one of several exclusive values.   Then maybe use "exclusive value"
* "Expanding arg references" section. I think it should include an example of the expansion and say example xxx is equivalent to example yyy
* "inicate" -> "indicate"

src/intent.html
```diff
@@ -370,6 +369,77 @@ <h3 id="intent_using">Using Intent Concepts and Properties</h3>
...
+    <dfn>self-properties</dfn>. The effect of any such property is
+    limited to the element (and its descendants) although the precise
+    effect depends on the definition of the property. So for
+    example:</p>
+    <ul>
+    <li>
+    <code>&lt;mrow
+    intent=":pause-medium">&lt;mi>x&lt;/mi>...&lt;/mrow></code>
+    would inicate there should be a pause in speech before the
+    <code>mrow</code> is procesed  (it would not cause a pause before
```
physikerwelt, line 427: )?

src/intent.html
```diff
@@ -370,6 +369,77 @@ <h3 id="intent_using">Using Intent Concepts and Properties</h3>
...
+    <i>intent</i> value starting with a colon, <code>:</code>) are known as
+    <dfn>self-properties</dfn>. The effect of any such property is
+    limited to the element (and its descendants) although the precise
+    effect depends on the definition of the property. So for
+    example:</p>
+    <ul>
+    <li>
+    <code>&lt;mrow
+    intent=":pause-medium">&lt;mi>x&lt;/mi>...&lt;/mrow></code>
+    would inicate there should be a pause in speech before the
```
physikerwelt, line 426: indicate?

src/intent.html
```diff
@@ -370,6 +369,77 @@ <h3 id="intent_using">Using Intent Concepts and Properties</h3>
...
+    <code>mrow</code> is procesed  (it would not cause a pause before
+    all descentant elements.
```
physikerwelt, line 428: descendant

src/intent.html
physikerwelt, line 427: processed

src/intent.html
```diff
@@ -370,6 +369,77 @@ <h3 id="intent_using">Using Intent Concepts and Properties</h3>
...
+    </li>
+    <li><code>&lt;mrow
+    intent=":literal">&lt;mi>x&lt;/mi>...&lt;/mrow></code>
+    would set the <i>parameter</i> corresponding to the default intent inference
+    rules to <code>literal</code> this would cause the intent for this
```
physikerwelt, line 433: new sentence

src/intent.html
```diff
@@ -370,6 +369,77 @@ <h3 id="intent_using">Using Intent Concepts and Properties</h3>
...
+    <code>mrow</code> including all its descendants to use the
```
physikerwelt, line 434: , including all its descendants,

src/intent.html
davidcarlisle: thanks. fixed in up coming commit final `)` added.

davidcarlisle:
> Some suggested edits:
> 
>     * ...system dependent however systems... => system dependent. However, systems
> 
>     * If multiple properties appear in a [self-property-list](...) or [term](...) then they  => add a comma before "then"
> 
>     * "_parameter_" is italicized, but there is no definition of what a parameter is. I think somewhere there should be a statement that some properties can have one of several exclusive values.   Then maybe use "exclusive value"
> 
>     * "Expanding arg references" section. I think it should include an example of the expansion and say example xxx is equivalent to example yyy
> 
>     * "inicate" -> "indicate"

Neil I did all these except the _parameter_ suggestion we can adjust wording later but I didn't want to wordsmith this at this point.
For the examples I took the whole block of examples from the issue. The bulk work of converting them from the markdown to the respec source in a folding examples section is done. The actual example text will need further edits (and probably fewer examples)

src/intent.html
davidcarlisle: done

src/intent.html
davidcarlisle: done

davidcarlisle:
sigh I merged without the above final updates I'll adjust

PR #547, opened 2025-11-06
remove decimalpoint from mstyle

NSoiffer:
Small fix for #1 -- remove `decimalpoint` from `mstyle`. This still need to be made a global attr to align with core.

Note: it also needs to be removed from RelaxNG.

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -4152,28 +4152,7 @@
    see <a href="#presm_lbattrs"></a>
   </td>
  </tr>
- 
- <tr>
- <td rowspan="2" class="attname"><span class="coreno">decimalpoint</span></td>
- <td><em>[=character=]</em></td>
- <td>.</td>
- </tr>
- 
- <tr>
-  <td colspan="2" class="attdesc">
-   Specifies the character used to determine the alignment point within
-   <a class="intref" href="#presm_mstack"><code class="element">mstack</code></a>
-   and
-   <a class="intref" href="#presm_mtable"><code class="element">mtable</code></a> columns
-   when the "decimalpoint" value is used to specify the alignment.
-   The default, ".", is the decimal separator used to separate the integral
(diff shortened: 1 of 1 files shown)


davidcarlisle:
decimalpoint isn't in core at all 

https://www.w3.org/TR/mathml-core/#global-attributes

so we need to add it to 

https://w3c.github.io/mathml/#presm_presatt

which I think should be part of this PR (I can fix up the schema separately in the mathml-schemas repo)

That would make `decimalpoint` the only attribute allowed on all presentation elements that's not inherited from mathml-core, but that's probably OK.

NSoiffer:
Not being in core is yet another reason why `decimalpoint` should be left off the global attrs -- we should be looking to minimize the differences. The two places that it is used are on `mstack` and in the `mtable` atts. Neither of these are implemented in core. `mstack` has some chance of being added in the future, but I suspect that adding alignment to `mtable` will get a lot of push-back because that's more the realm of CSS.

davidcarlisle:
> Not being in core is yet another reason why `decimalpoint` should be left off the global attrs -- we should be looking to minimize the differences. The two places that it is used are on `mstack` and in the `mtable` atts. Neither of these are implemented in core. `mstack` has some chance of being added in the future, but I suspect that adding alignment to `mtable` will get a lot of push-back because that's more the realm of CSS.

I'm not sure what you propose here: if we remove it from mstyle and don't add it back somewhere the feature has gone so we need to remove `decimalpoint` as an attribute value in mtable and mstack unless we are only going to support `.` which probably isn't acceptable. If it's not allowed on all presentation elements I think it needs to be at least allowed where attributes that take `decimalpoint` as a value are allowed, so `mstack` and `mtable` (and `mtr` and `mtd`)

NSoiffer:
David and I discussed this. I'm going to close this and resubmit a new PR that breaks apart the two PRs and changes the `decimalpoint` change.

Issue #1, opened 2019-02-14
Simplification of the mstyle element


fred-wang:
This proposal is to simplify mstyle so that it only supports the following attributes:
* displaystyle
* dir
* mathsize
* mathbackground
* mathcolor
* mathvariant
* scriptlevel

And maybe:
* scriptsizemultiplier
* scriptminsize

From the Math WG mailing list:
https://lists.w3.org/Archives/Public/www-math/2016Jul/0024.html

From the appendix of the original "MathML in HTML5" document:
"The mstyle element duplicates the CSS inheritance mechanism in a CSS-incompatible way and has many exceptions to workaround the fact that attribute names may be used for different purposes. In practice, many of its attributes are not useful and never used. It is a burden for implementers since they essentially have to reimplement a specific “attribute” inheritance mechanism to support the general case even if the most prominent attributes have obvious mapping to CSS. It is also a performance issue to perform the rendering and keep it up-to-date since the rendering on any node may depend on its mstyle ancestors. In this document, it instead suggested to keep only attributes that are compatible with CSS and five new CSS properties are introduced in section 2.3.1. These attributes actually correspond to what is used in practice. "

edit:
This is the proposal for CSS attributes that cannot be mapped to style yet: https://github.com/mathml-refresh/mathml-css-proposals#mathml-css-proposals

physikerwelt:
This might be a stupid question, but why do we need 
* dir -> direction
* mathsize -> font-size
* mathbackground -> background-color
* mathcolor -> color
can that not implemented using the style attribute?

fred-wang:
Yes, they can be (and are) implemented using CSS style.
Whether we want to remove/deprecate these MathML attributes (and hence the whole mstyle element) is another question, but probably they have already been used a lot in existing documents. This is not really the case of attributes not mentioned in https://github.com/mathml-refresh/mathml/issues/1#issue-410409466 , I think in a previous discussion on Math WG only one instance of another attribute (linethickness) was used in the MathML 3 test suite.

fred-wang:
Incidentally, for attributes without CSS equivalent, there is https://github.com/mathml-refresh/mathml-css-proposals#mathml-css-proposals

@physikerwelt I think in general it would be good to gather usage metrics of elements/attributes that are proposed for deprecation/removal. Maybe you can do that for wikipedia.

davidcarlisle:
> 
> 
> This might be a stupid question, but why do we need
> 
>     * dir -> direction
> 
[...]

PR #549, opened 2025-11-06
Space like

### LINKED ISSUE #548: Modify the space-like list to include mtext only if contains only Unicode spacing chars
At the 30 Oct meeting, we discussed https://github.com/w3c/mathml-core/issues/286. The conclusion was both full and core should align on a definition of space-like elements. `mtext` should be space-like only if it contains only Unicode spacing characters (but not the space, cr, or nl chars).

Here's a list of the characters (comes from AI, so needs to be double checked):
```
&#x00A0;,&#x1680;,&#x2000;,&#x2001;,&#x2002;,&#x2003;,&#x2004;,&#x2005;,&#x2006;,&#x2007;,&#x2008;,&#x2009;,&#x200A;,&#x202F;,&#x205F;,&#x3000;,&#x200B;,&#x200C;,&#x200D;,&#x2060;,&#xFEFF;
```

NSoiffer:
Make `mtext` be more restrictive as to when it is space-like as per the 28 Oct, 2025 meeting.

Fixes #548

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -4152,28 +4152,7 @@
    see <a href="#presm_lbattrs"></a>
   </td>
  </tr>
- 
- <tr>
- <td rowspan="2" class="attname"><span class="coreno">decimalpoint</span></td>
- <td><em>[=character=]</em></td>
- <td>.</td>
- </tr>
- 
- <tr>
-  <td colspan="2" class="attdesc">
-   Specifies the character used to determine the alignment point within
-   <a class="intref" href="#presm_mstack"><code class="element">mstack</code></a>
-   and
-   <a class="intref" href="#presm_mtable"><code class="element">mtable</code></a> columns
-   when the "decimalpoint" value is used to specify the alignment.
-   The default, ".", is the decimal separator used to separate the integral
-   and decimal fractional parts of floating point numbers in many countries.
-   (See <a href="#presm_elementary"></a> and <a href="#presm_malign"></a>).
- </td>
- </tr>
- </tbody>
- </table>
- 
+  
  <p>If <code class="attribute">scriptlevel</code> is changed incrementally by an
  <code class="element">mstyle</code> element that also sets certain other
  attributes, the overall effect of the changes may depend on the order
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -3272,8 +3272,8 @@
  <ul>
  
   <li>
-   <p>an <code class="element">mtext</code> element all of whose characters are
-    Unicode spacing characters except U+0020 (space);</p> </li>
+   <p>an <code class="element">mtext</code> element that is empty or that has only text content which contains solely
+    Unicode spacing characters;</p> </li>
  
   <li>
    <p>an <code class="element">mspace</code>,
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -3272,7 +3272,11 @@
  <ul>
  
   <li>
-   <p>an <code class="element">mtext</code>, <code class="element">mspace</code>,
+   <p>an <code class="element">mtext</code> element all of whose characters are
+    Unicode spacing characters except U+0020 (space);</p> </li>
+ 
+  <li>
+   <p>an <code class="element">mspace</code>,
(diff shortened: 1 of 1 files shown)


src/presentation-markup.html
```diff
@@ -3272,7 +3272,11 @@ <h5 id="presm_space_like">Definition of space-like elements</h5>
  <ul>
  
   <li>
-   <p>an <code class="element">mtext</code>, <code class="element">mspace</code>,
+   <p>an <code class="element">mtext</code> element all of whose characters are
+    Unicode spacing characters except U+0020 (space);</p> </li>
```
davidcarlisle: why exclude space? That seems a bit weird, you are saying `<mtext> </mtext>` is not space-like?

NSoiffer: I excluded it because it is stripped from being displayed (or it should be). That would leave an empty `mtext`. I suppose that's ok, just a little weird because then it doesn't display as white space. But I suppose that is also true of `mspace` with `width='0' height='0' depth='0'`.

NSoiffer:
I changed the wording. I only tried to make it clear that it is space-like if there is only textual content and that content only has space chars. The wording is a little awkward because I didn't want to repeat "only" and "that" several times.

src/presentation-markup.html
```diff
@@ -4152,28 +4156,7 @@ <h5 id="presm_mstyle_attrs">Attributes</h5>
    see <a href="#presm_lbattrs"></a>
   </td>
  </tr>
- 
- <tr>
- <td rowspan="2" class="attname"><span class="coreno">decimalpoint</span></td>
```
davidcarlisle, line 4157: was this deletion intentended in this PR?
If we delete it from the mstyle-attributes don't we need to add it as a common presentation  mathml, or or mrow attribute, as it's used elsewhere (also the schema would need an update to match)

src/presentation-markup.html
NSoiffer: > was this deletion intended in this PR?

No. I checked out a separate branch and put the change there, but I guess I messed up the push.

See https://github.com/w3c/mathml/issues/1#issuecomment-3033059347 for agreement to drop it from mstyle. It doesn't make much sense to inherit it, similar to other non-inherited attributes (e.g., `linethickness`).

davidcarlisle: Currently `decimalpoint` is _only_ allowed on `mstyle` so if you drop it from there it is not allowed anywhere, it's not just a matter of inheritance.  I note the comment there does say it can be dropped which I suppose makes sense if we don't support this. Although if we allowed the syntax it could be poyfilled. (but this PR is wrong place for this discussion) I suggest we merge the mtext change and move this to a separate PR.

  the mtext change makes sense but does it rely on synchronisation with mathml core? If core implementations don't want to inspect element content to make this distinction do we still do it here and polyfill?

NSoiffer: I just realized the source of confusion: we have an attribute on mstack called "decimalpoint" which specifies that alignment should be on a decimal point. But we also have the mstyle attr that is also called "decimalpoint". That specifies the character to use for a decimal point. That was a poor choice. It should probably have been called "decimalchar" or "decimalcharacter" (or better, something with a hyphen, but don't do that in MathML). Too late to change the name when moving it to a global attr? It probably has minimal usage due to the lack of elementary math implementations. In any case, "decimalpoint" to specify the char to use does make sense to be global as it would be really weird to have more than one value in an expr, so setting on `math` or `stackalign` makes sense.

I was surprised to see that mtable's `columnalign` only has left, right, and center. No decimal point. It appears we left that alignment to groupalign, but that's gone.

davidcarlisle: no there is only the attribute on mstyle. mstack and mtable take `decimalpoint` as a _value_ like `left` and `right` to specify an alignment on decimals using the character specified on a surrounding mstyle, or `.` by default.  there is an mtable example at https://w3c.github.io/mathml/#presm_ex_alignment_1  `<mtable groupalign="{decimalpoint left left decimalpoint left left decimalpoint}">`  That is essentially the same as MathML3.

NSoiffer: I'll start a new PR.

NSoiffer:
I'll close and start again

PR #550, opened 2025-11-13
change the meaning of space-like

NSoiffer:
Clean PR for the change to space-like

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -3273,7 +3273,7 @@
  
   <li>
    <p>an <code class="element">mtext</code> element all of whose characters are
-    Unicode spacing characters except U+0020 (space);</p> </li>
+    Unicode spacing characters;</p> </li>
  
   <li>
    <p>an <code class="element">mspace</code>,
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -3272,7 +3272,7 @@
  <ul>
  
   <li>
-   <p>an <code class="element">mtext</code> element all of whose characters are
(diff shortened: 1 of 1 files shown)


davidcarlisle (changes requested):
as commented previously I don't think it should say " except U+0020 (space);"  also the wording may need adjusting depending on how closely mathml-core aligns with this

src/presentation-markup.html
```diff
@@ -3272,7 +3272,11 @@ <h5 id="presm_space_like">Definition of space-like elements</h5>
  <ul>
  
   <li>
-   <p>an <code class="element">mtext</code>, <code class="element">mspace</code>,
+   <p>an <code class="element">mtext</code> element all of whose characters are
+    Unicode spacing characters except U+0020 (space);</p> </li>
```
davidcarlisle:
```suggestion
    Unicode spacing characters;</p> </li>
```

davidcarlisle (changes requested):

src/presentation-markup.html
```diff
@@ -3272,7 +3272,11 @@ <h5 id="presm_space_like">Definition of space-like elements</h5>
  <ul>
  
   <li>
-   <p>an <code class="element">mtext</code>, <code class="element">mspace</code>,
+   <p>an <code class="element">mtext</code> element all of whose characters are
```
davidcarlisle:
```suggestion
   <p>an <code class="element">mtext</code> element that is empty or all of whose characters are
```

davidcarlisle (approved):

PR #552, opened 2025-11-13
add table of suggested chars for accents/lines

NSoiffer:
This still needs work/discussion. Implementations are all over the place (or don't work for any character). Not sure what we should do.

### DIFF
--- a/src/character-set.html
+++ b/src/character-set.html
@@ -526,6 +526,102 @@
 
   </section>
 
+ <section>
+   <h4 id="accent-characters">Accent Characters</h4>
+
+   <p> There are a large number of look-alike characters that might be used to for
+    accents or diacritical marks above or below a base element.
+    However, fonts have limitations as to which of these can stretch and which have
+    proper positioning when used as accents. This section has recommendations
+    for which characters to use as accents in MathML. It is important to note that
+    only a limited number of fonts support proper mathematical layout.
+    Browsers and likely other software currently require the use of fonts with an OpenType MATH
+    <table id="suggested-accent-chars">
+
+    <thead>
+
+     <tr>
+      <td>Unicode code point</td>
+      <td>Unicode name</td>
+      <td>Example</td>
+     </tr>
+    </thead>
+
+    <tbody>
+     <tr>
+      <td>U+0302</td>
+      <td>Hat (stretches)</td>
+      <td><math></math><mover><mi>xyz</mi><mo>&#x302;</mo></mover><math></math></td>
+     </tr>
+
+     <tr>
+      <td>U+02D8</td>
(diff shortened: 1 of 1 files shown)


src/character-set.html
```diff
@@ -526,6 +526,102 @@ <h4 id="chars_comb-chars">Combining Characters</h4>
...
+ <section>
+   <h4 id="accent-characters">Accent Characters</h4>
+
+   <p> There are a large number of look-alike characters that might be used to for
+    accents or diacritical marks above or below a base element.
+    However, fonts have limitations as to which of these can stretch and which have
+    proper positioning when used as accents. This section has recommendations
+    for which characters to use as accents in MathML. It is important to note that
+    only a limited number of fonts support proper mathematical layout.
+    Browsers and likely other software currently require the use of fonts with an OpenType MATH
```
davidcarlisle:
```suggestion
    Browsers and likely other software currently require the use of fonts with an OpenType MATH
    table.
```

src/character-set.html
```diff
@@ -526,6 +526,226 @@ <h4 id="chars_comb-chars">Combining Characters</h4>
...
+			<td><math><mover accent="true"><mrow><mi>x</mi><mo>+</mo><mi>y</mi></mrow><mo stretchy='true'>&#x007E;</mo></mover></math></td>
+			 <td>U+0303 doesn't stretch; U+007E doesn't stretch in Chrome</td>
+		</tr>
+    <tr>
+			<td>U+005F</td>
+			<td>lowline (mover)</td>
+			<td>&#x005F;</td>
+			<td><math><mover accent="true"><mi>x</mi><mo>&#x005F;</mo></mover></math></td>
+			<td><math><mover accent="true"><mrow><mi>x</mi><mo>+</mo><mi>y</mi></mrow><mo stretchy='true'>&#x005F;</mo></mover></math></td>
+			<td>U+00AF (overline) doesn't stretch in Chrome and displays poorly in Safari. U+0304 and U+0305 don't stretch anywhere</td>
```
davidcarlisle, line 590: I don't think we can be making comparative judgements of specific implementations in the spec, I think we should just have a paragraph describing the issue. We can link to appendix B.3 of mathml-core (assuming we land the PR to update that and also, we could link  to MathFonts.github.io accents test to tell people to see which characters stretch in current implementations.

davidcarlisle (changes requested):

davidcarlisle:
MathML Core B3 has been updated so now isn't totally wrong.

https://w3c.github.io/mathml-core/#combining-character-equivalences

There is more that could be said there but I think we can finalize things in the main spec by referencing that now.

PR #553, opened 2025-11-16
Move mstyle attrs to attrs common to all presentation elements

NSoiffer:
This mostly addresses #1.

### DIFF
--- a/src/fundamentals.html
+++ b/src/fundamentals.html
@@ -752,8 +752,9 @@
   <section>
    <h4 id="interf_toplevel_atts">Attributes</h4>
 
-   <p>The <code class="element">math</code> element accepts any of the attributes that can be set on
-   <a href="#presm_mstyle"></a>, including the common attributes
+   <p>The <code class="element">math</code> element accepts any of the common presentation attributess specified in
+   <a id="presm_presatt">Mathematics attributes common to presentation elements</a>,
+   along with the common attributes
    specified in <a href="#fund_globatt"></a>.
    In particular, it accepts the <code class="attribute">dir</code> attribute for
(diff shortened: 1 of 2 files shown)


src/fundamentals.html
```diff
@@ -516,7 +516,7 @@ <h5 id="fund_defaults">Default values of attributes</h5>
    <h4 id="fund_globatt">Attributes Shared by all MathML Elements</h4>
 
    <p>In addition to the attributes described specifically for each element,
-   the attributes in the following table are allowed on every MathML element.
+   the HTML attributes in the following table are allowed on every MathML element.
```
davidcarlisle:
```suggestion
   the attributes in the following table are allowed on every MathML element.
```

I wouldn't add HTML here I think it's confusing. xref isn't HTML at all and MatHMl's href is a global attribute so not really like HTML/a's href which is just on a (and similar elements)

src/presentation-markup.html
```diff
@@ -1063,8 +1063,90 @@ <h5 id="presm_presatt_core">MathML Core Attributes</h5>
        background color, if any, used in the current rendering context to show through.
       </td>
      </tr>
+
```
davidcarlisle: If you add to this table rather than make a new one, then need to restructure the leading  paragraph

> These attributes include the following two attributes that are primarily intended for visual media. ....

as it now has more than two entries and not all for same usage.

davidcarlisle (changes requested):
looks good, couple of small comments made inline

NSoiffer:
@davidcarlisle: I found a few mistakes (displaystyle and scriptlevel got chopped) and also fixed up some references to mstyle. Please review.

PR #556, opened 2025-12-10
Make attributes with enumerated string values be ASCII case-insensitive 

dginev:
Addresses the "enumareted string values" piece of #178 .

I haven't seen the tables rendered, there is a chance they got quite verbose. We could compress to a column with a checkmark if preferred, I tried the low-tech edit first.

### DIFF
--- a/src/fundamentals.html
+++ b/src/fundamentals.html
@@ -339,6 +339,9 @@
     a single whitespace character for separating parts of a value.
     Moreover, leading and trailing whitespace in attribute values should be avoided.</p>
 
+    <p>For compatibility with HTML, attributes defined with an enumerated list of string values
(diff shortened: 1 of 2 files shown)


NSoiffer (commented):
I'm not a fan of replicating `Comparisons are <a data-cite="INFRA#ascii-case-insensitive">ASCII case-insensitive</a>, where all matching strings are allowed as values.` However, core does this and so I guess MathML 4 should do the same.

I believe that Content MathML should be consistent in this respect also. Do values of "type" or any other attr need this wording?

davidcarlisle:
A rendering of the spec based on this branch is online at

https://davidcarlisle.github.io/mathml/spec.html#presm

davidcarlisle (approved):

PR #557, opened 2025-12-11
decimalpoint -> decimalcharacter

NSoiffer:


davidcarlisle (approved):
approve, just noting a matching change in the mathml-schema repo is needed

Issue #555, opened 2025-12-10
Add <a> to spec


NSoiffer:
As agreed to in the Dec 4, 2025 meeting, `<a>` should be added to the spec. It should be considered to be identical to `<mrow>` with the comment that in core, the `href` attr works.

davidcarlisle:
This has been added by PR #558 however the added example prevents the publishing to TR space so currently it is only in github pages.

```
  💥 Echidna publish has failed.
  Specberus Errors:
    [6177@171] Element “a” not allowed as child of element “math” in this context. (Suppressing further errors from this subtree.)<br><a href="https://validator.w3.org/nu/?doc=https://mimas.w3.org/6f0defaf-de76-4562-bb78-26613c8f4499/Overview.html">HTML validator</a>
      { name: 'validation.html', section: 'format', rule: 'validHTML' }

```

@NSoiffer   I could mark the example as not core, then the build would not inline the example, which might be an appropriate quick fix to get the update to TR space

davidcarlisle:
Ok finally it passes echidna validation, basically the example displays as `<a href` but uses mrow and onclick events in the inlined display.

davidcarlisle:
The schema has been updated to match this issue and already shows at the live respec version

https://w3c.github.io/mathml/spec.html#parsing_rnc_core

it won't show at the static build version until the next build.

`<a>` added to mathml-core with the same attributes as in HTML, and in addition those attribute added to `mrow` in full.

https://github.com/w3c/mathml-schema/commit/d086a424774e230908a7c683a5dc2ac93ac7ba7c

davidcarlisle:
`<a>` is added to mathml4 and core

PR #558, opened 2025-12-18
a element for #555

davidcarlisle:
For issue #555 This adds `<a>` as an alias for `<mrow>` in mathml full (with a reference to `<a href` working in mathml core 

matching PR will be needed in mathML-Core and MathML-Schema


NSoiffer:
Approved pending the addition of an example (as per the meeting today).

PR #562, opened 2026-01-09
A element further adjustments

davidcarlisle:
This 

* deletes the description of `dir` as an additonal mrow-specific attribute (as it is already allowed on all presentation elements, inherited from mathml-core)

* updates the linking section in chapter 7 (which was explicitly describing MathML **3**)

* documents `<mrow>`  and `<a>` as taking attributes such as `target=_blank` from HTML `<a>`

(this last part may need some adjustment depending on how exactly `<a>` gets added to mathml-core, but this at least gets the current drafts aligned.


NSoiffer (approved):
Looks good.

PR #564, opened 2026-01-15
rewrote some parts of mstyle

NSoiffer:
Hopefully clarifies when attribute values are applied to children elements

### DIFF
--- a/src/presentation-markup.html
+++ b/src/presentation-markup.html
@@ -4207,24 +4207,31 @@
  <code class="attribute">mathsize</code>, which can only be set on token elements (or on
  <code class="element">mstyle</code> itself).</p>
  
- <p>MathML specifies that when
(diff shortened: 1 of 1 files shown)


davidcarlisle (changes requested):
I think the changed text is fine but I think we should say more about the differences to core.

in core mstyle is an alias for mrow essentially and just allows the attributes allowed on on all presentation elements. so in the first paragraph

> The mstyle element is used to make style changes that affect the rendering of its contents. As a presentation element, it accepts the attributes described in [3.1.9 Attributes common to presentation elements](https://www.w3.org/TR/mathml4/#presm_presatt). **Additionally, it can be given any attribute accepted by any other presentation element, except for the attributes described below.**

The last sentence highlighted above is an additional non-core feature and the remaining descriptions of attributes correspond to these non-core attributes and how they are interpreted.  So I think we should say that and also point to a polyfill that handles mstyle (which I'm not sure exists yet?)

davidcarlisle (approved):

PR #567, opened 2026-03-11
Proofread Chapter 5, Intent

dginev:
A proofreading pass over Chapter 5.

I've had help from Claude Opus 4.6 "heavy effort" in editing here and may do a more human pass a little later.


davidcarlisle (approved):
looks good thanks

PR #571, opened 2026-03-26
Proofread spec text

dginev:
Proofread rest-of-spec (skipped Chapter 5 and all Content MathML text)

Mostly managed by Claude Opus 4.6 "heavy effort".

Whitespace-free diff should be available at:
https://github.com/w3c/mathml/pull/571/changes?w=1

### DIFF
--- a/src/accessibility.html
+++ b/src/accessibility.html
@@ -112,7 +112,7 @@ classroom for young students.</p>
 <p>Braille output. The tactile rendering of mathematical expressions
 in braille is a very important use case. For someone who is blind,
 interpreting mathematics through auditory rendering alone is a
-cognitive taxing experience except for the most basic expressions. And
+cognitively taxing experience except for the most basic expressions. And
 for a deafblind user, auditory renderings are completely
 inaccessible. Several math braille codes are in common use globally,
 such as the Nemeth braille code, UEB Technical, German braille
@@ -147,7 +147,7 @@ for individuals who are deaf or hard of hearing would be the ability
 to provide input to automated signing avatars. Automated signing
 avatar technology which generates American Sign Language has already
 been applied to elementary level mathematics <span class="note">add citation</span>. Sign
-languages vary by county (and sometimes locality) and are not simply
+languages vary by country (and sometimes locality) and are not simply
 "word to sign" translations, as sign language has its own grammar, so
 being able to access the underlying tree structure of mathematical
 expressions as can be done with MathML will provide the potential for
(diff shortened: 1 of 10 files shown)


src/accessibility.html
```diff
@@ -225,7 +225,7 @@ <h5 id="acc_at">Accessibility tree</h5>
 this method. For instance, the entire spoken text of the expression
 must be given in the tag, even if the author is only concerned about
 one small portion. Further, <code class="attribute">alttext</code> is
-limited to plain text, so speech queues such as pausing and pitch
+limited to plain text, so speech cues such as pausing and pitch
```
davidcarlisle, line 228:
```suggestion
limited to plain text, so speech queues such as pausing and pitch
```

src/fundamentals.html
```diff
@@ -903,7 +903,7 @@ <h4 id="interf_toplevel_atts">Attributes</h4>
        A positive value of <code class="attribute">altimg-valign</code> shifts the bottom of the image above the
        current baseline, while a negative value lowers it.
        The keyword "top" aligns the top of the image with the top of adjacent inline material;
-       "center" aligns the middle of the image to the middle of adjacent material;
+       "middle" aligns the middle of the image to the middle of adjacent material;
```
davidcarlisle, line 906:
```suggestion
       "center" aligns the middle of the image to the middle of adjacent material;
```

davidcarlisle (changes requested):
Looks good I just backed out two changes see inline comments

src/fundamentals.html
dginev: If "center" is the correct altimg-valign value, I can edit the triple at the top to include it instead of "middle"?

Screenshot From 2026-03-26 06-19-31

src/accessibility.html
dginev: I think the hand-raising in zoom can be called a "speech queue", but the pausing and pitch adjustment are typically called a "speech cue" ?

NSoiffer: That's almost certainly my typo -- I've caught myself doing that many times. "cues" is correct.

src/changes.html
```diff
@@ -41,7 +41,7 @@ <h4 id="chg_fund">Changes to <a href="#fund"></a></h4>
     The `mathml4-legacy` schema makes these valid if needed for legacy applications.</li>
 
     <li>Remove the <code class="attribute">other</code> attribute.
-    This have been deprecated
+    This had been deprecated
```
NSoiffer:
```suggestion
    This has been deprecated
```

src/fundamentals.html
NSoiffer: Looking back to MathML 3, it has always been `middle`.  So +1 for Claude, -1 for David.

src/changes.html
NSoiffer: @davidcarlisle or @dginev: do you agree with my suggestion?

A lot of the changes were meaningless extra space cleanups, but I'm impressed it caught some actual errors in the spec that weren't just spelling/grammar changes!

dginev: has/had both read good to me here, if subtly different in meaning. Sure, whichever you prefer.

Issue #560, opened 2026-01-04
mathvariant attribute enumeration is missing chancery and roundhand


MurrayIII:
In Section 3.2.3.2 Attributes for mathvariant, the variants roundhand and chancery are missing.

davidcarlisle:
meeting of 19th Feb  agreed to add this to mathml4 spec

physikerwelt:
See https://github.com/mathjax/MathJax/issues/3292 for the related MathJax problem

davidcarlisle:
decided to close this as mathvariant is not in core (amongst other issues

will add a note that variant selector FE00 FE01 may be used to disambiguate script....

davidcarlisle:
closed by #573

PR #573, opened 2026-05-12
describe script variants for #560

davidcarlisle:
Document the use of U+FE00 and U+FE01 to select script variants as agreed in issue #560


NSoiffer:
I looks ok to me, but I defer to @MurrayIII as to whether it is technically correct.

MurrayIII (approved):
Looks good to me.

Issue #522, opened 2024-12-15
nested math in token elements.


davidcarlisle:
The following issue arose while working on examples using MathML for accessibility tagging in PDF, looking at typical tex cases
such as a display containing nested text such as

```
\[
f(n) = \begin{cases}
        1 & \text{if $n$ is prime} \\
        0 & \text{otherwise}
       \end{cases}
\]
```

$$
f(n) =\begin{cases}
    1 & \text{if $n$ is prime} \\
     0 & \text{otherwise}
    \end{cases}
$$

MathML3 and 4 allow HTML elements in presentation token elements when used in html, deferring to the HTML(5) spec for exactly which elements are allowed.

HTML specifies the phrasing content is allowed 

https://html.spec.whatwg.org/#phrasing-content-2

The schema used with mathml3 and the current draft mathml-core schema interpreted this by extending the content model of token elements to take text or html, this means that to make a valid version of the above with a nested `$n$`
 in `<mtext>` you need an intervening html `<span>`

```
<mtext><span>if <math><mi>n</mi></math> is prime</span></mtext>
```

However it was pointed out, correctly I think, that phrasing content at the above link, explicitly lists mathml namespace `<math>` and svg  namespace `<svg>` as phrasing content, so that they should be valid children of `<mtext>` and that current browsers do all implement that correctly and this mathml does render as intended.

```
<mtext>if <math><mi>n</mi></math> is prime</mtext>
```

This could be fixed in the draft core schema by changing the `textorHTML` production

https://w3c.github.io/mathml/#parsing_textorHTML

to include `| math | svg`

and possibly to highlight this in the section on mathml in html at 

https://w3c.github.io/mathml/#interf_html

Given that HTML has always specified phrasing content this way and that it is implemented this way in browsers, I think changing the schema is essentially an editorial correction (especially as only draft versions of it have been published so far anyway)

Apart from allowing simpler markup and matching current browser implementations, allowing math directly under mtext means that the common case of nested math which _just_ has text and math such as `\text{when $x<0$}` can be modelled just using a single, mathml, namespace. You only need to bring in the additional html namespace to model nested textual markup such as `\text{something \textbf{bold} about $x$}` which is far less common.

So while arguing that this could be viewed as an editiorial correction I think it's a big enough change that it should be logged as an issue, hence this issue.

physikerwelt:
[...]

PR #574, opened 2026-05-13
math in mtext example for #522

davidcarlisle:
As agreed I think in issue #522 this adds an example of `<math>` in `<mtext>` in the existing section on embedding in token elements, and points out that while technically an extension it's allowed in MathML-in-HTML and MathML-Core and any embeddings based on MathML Core. An additional note is added in the main section introducing `<mtext>` referencing this.


NSoiffer:
Looks ok. I can't tell what the example looks like rendered, but I'll find out soon...

Issue #576, opened 2026-05-18
Privacy Review - MathML 4


gbshankar:
# Scope 

MathML4 delta from MathML Core. MathML Core's prior PING review ([issue \#130](https://github.com/w3cping/privacy-request/issues/130)) is taken as a baseline where applicable, but this review covers MathML Full-specific features that Core intentionally omits.

# Overall assessment

**No blocking privacy concerns**. For features shared with MathML Core, the baseline Core review stands. MathML 4’s Privacy and Security Considerations sections (§D.4 and §D.5) are nevertheless insufficient because they largely defer to Core, while MathML Full introduces several features outside Core’s scope — specifically `intent`, the `annotation src` reference mechanism, `mglyph`, broad `href` on all elements, and Content MathML semantic identifiers — each of which warrants explicit treatment. The primary new privacy risk is misuse of `intent` as a hidden alternate content channel, potentially enabling assistive-technology-use detection. All findings appear addressable with targeted additions to the specification text.

---

## Finding 1 — §D.4 and §D.5 must not simply defer to MathML Core

MathML Core’s privacy and security sections address the subset of MathML defined by Core, including Core-specific rendering, layout, font, link, and embedding risks. MathML 4 Full introduces features outside Core’s scope.Simply deferring to MathML Core leaves several MathML Full-specific features without explicit privacy or security analysis, including `intent` processing, `annotation src` external references, `mglyph` resource loading, broad `href` support on all elements, and Content MathML semantic identifiers.

*Request:* MathML 4 should include Privacy and Security Considerations sections that explicitly address MathML Full-specific features rather than relying on the Core review alone.

---

## Finding 2 — `href` on all MathML elements reintroduces link-model risks outside Core

MathML Core did not retain MathML3’s broad `href/xlink:href` model. MathML 4 Full reinstates `href` on all MathML elements, including invisible and nested elements that may not fit the ordinary HTML link model. The specification should clarify activation behavior and privacy protections for these cases, including visited-link handling.

*Requested addition to §D.5:*

In web contexts, MathML `href` must not create link, navigation, URL-scheme, referrer, script-execution, download, or target-handling capabilities beyond those allowed by the host environment’s ordinary link model. The specification should also clarify safe behavior for `href` on non-rendered elements and for nested MathML links, since these cases may not map cleanly to ordinary visible links.

---

## Finding 3 — AT-use detection via intent divergent content (primary new privacy concern)

[...]

PR #581, opened 2026-05-30
update for #576

davidcarlisle:
This implements in full (I hope) the suggestions from #576

### DIFF
--- a/src/conformance.html
+++ b/src/conformance.html
@@ -316,16 +316,128 @@
 
       <section>
       <h3 id="privacy">Privacy Considerations</h3>
-      <p>Web platform implementations of MathML should implement [[MathML-Core]],
-	and so the <a href="https://www.w3.org/TR/mathml-core/#privacy-considerations">Privacy Considerations</a> specified there apply.</p>
-    </section>
+  <p>Web platform implementations of MathML should implement [[MathML-Core]],
+  and so the <a href="https://www.w3.org/TR/mathml-core/privacy-considerations">Privacy Considerations</a>  specified there apply in addition to the
+  considerations listed below.</p>
+  
+  <section>
+    <h4 id="privacy_intent">intent</h4>
+    <p>Authors must not use <code class="attribute">intent</code> to
+    convey hidden instructions, behavioral probes, tracking tokens, or
+    content that materially differs from the visible mathematical
+    expression. intent should be used only to disambiguate or improve
+    narration/navigation of the same mathematical content.</p>
     
+    <p>User agents should not expose to page script any signal
+    indicating whether, how, or by whom <code
+    class="attribute">intent</code> was consumed by assistive
+    technology.</p>
+
+
+    <p>The <code class="attribute">intent</code> attribute provides an
+    author-supplied semantic layer intended to improve mathematical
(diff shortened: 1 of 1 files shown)


src/conformance.html
```diff
@@ -316,16 +316,128 @@ <h4 id="interf_unspecified">Attributes for unspecified data</h4>
...
+  <p>Web platform implementations of MathML should implement [[MathML-Core]],
+  and so the <a href="https://www.w3.org/TR/mathml-core/privacy-considerations">Privacy Considerations</a>  specified there apply in addition to the
+  considerations listed below.</p>
+  
+  <section>
+    <h4 id="privacy_intent">intent</h4>
+    <p>Authors must not use <code class="attribute">intent</code> to
+    convey hidden instructions, behavioral probes, tracking tokens, or
+    content that materially differs from the visible mathematical
+    expression. intent should be used only to disambiguate or improve
```
NSoiffer: <code class="attribute">intent</code>

src/conformance.html
```diff
@@ -316,16 +316,128 @@ <h4 id="interf_unspecified">Attributes for unspecified data</h4>
...
+  </section>
+</section>
 
       <section>
       <h3 id="security">Security Considerations</h3>
       <p>Web platform implementations of MathML should implement [[MathML-Core]],
-	and so the <a href="https://www.w3.org/TR/mathml-core/#security-considerations">Security Considerations</a> specified there apply.</p>
+	and so the <a href="https://www.w3.org/TR/mathml-core/#security-considerations">Security Considerations</a>  specified there apply in addition to the
+  considerations listed below.</p>
       <p>In some situations, MathML expressions can be parsed as XML. The security considerations of XML parsing apply then as explained in [[?RFC7303]].</p>
```
NSoiffer, line 382: Is [[?RFC7303]] correct or is the "?" someone guessing as to the reference? It references [XML Media Types](https://www.rfc-editor.org/rfc/rfc7303), which seems wrong.

NSoiffer:
I think we also were going to do a small update to the intent section. Should that be a separate PR?

src/conformance.html
davidcarlisle: df41e68

src/conformance.html
davidcarlisle: It's respec markup I forget exactly what the  `?` does but we use it quite a few times throughout the spec It affects the form of the generated bibliography. That part was in the existing document not added by this PR @NSoiffer

davidcarlisle:
> I think we also were going to do a small update to the intent section. Should that be a separate PR?

@NSoiffer  er.. I forget, do you want to make a PR? It's probably easier than reminding me what I was supposed to add.

src/conformance.html
NSoiffer: Do you think the reference to media types is correct (in regards to XML parsing security considerations)?

davidcarlisle: @NSoiffer  It was suggested by @polx  3c8363f4d42c24607b0253476f40490d076f47bb so been there since 2022

davidcarlisle:
WG agreed to merge this

PR #583, opened 2026-07-30
Fix ReSpec build and Pages publishing

korrayt:
## Summary

- Use the default export exposed by `mathml-polyfills` so the ReSpec build can load `_MathTransforms`.
- Keep the nested-`math` presentation example as source-only markup instead of wrapping it in another generated outer `math` element.
- Let `w3c/spec-prod` manage the `gh-pages` branch after a successful build, and correct the W3C decision URL typo.

## Root cause

The latest CI build imports `_MathTransforms` as a named export, while `all-polyfills.js` exposes it as the default export. The `showcore` post-processing hook also wraps MathML examples in an outer `math` element; that made the nested-`math` example produce invalid generated markup. The workflow reset `gh-pages` before the build, so a failed build could leave the Pages branch without the generated document.

## Validation

- ReSpec 37.2.0 local build completed successfully (exit code 0).
- Generated HTML contained no nested real `math` elements.
- Workflow YAML parsed successfully with Prettier's YAML parser.
- `git diff --check` passed.

Related to #522.


.github/workflows/auto-publish.yml
```diff
@@ -12,12 +12,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
      - uses: actions/checkout@v5
-     - run: |
```
davidcarlisle, line 15: Currently all development is in main branch but we need to put the built documents on to gh-pages
If I'm reading this diff correctly you are making gh-pages the source branch? I'm not sure we want to change that at this time (although I can believe this step has issues, it has basically been working for some years?)
or are you saying spec-prod can handle that on its own now, I should check....

